PR #43, opened 2023-07-17
B/image review

uglrl:
## Backend
- [ ] Your code is **documented** using doc comments.
- [ ] You added appropriate **logging** in your code.
- [ ] You added **unit tests** to all your implemented functions.
- [ ] You ran `cargo format` to format your code.
- [ ] You added the **class descriptions** of the classes you implemented in the **latex document** on overleaf.


backend/src/layer/logic/image_review/image_reviewer.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    async fn review_images(&self, images: Result<Vec<Image>>) {
+        if let Err(error) = self.try_review_images(images).await {
+            warn!("{error:#?}");
+        }
+    }
+
+    async fn try_review_images(&self, images: Result<Vec<Image>>) -> ReviewerResult<()> {
+        let images = images.map_err(|_e| ReviewerError::ImageGetError)?;
```
worldofjoni, line 90: we dont want to discard the data error as we lose context this way. We shoud rather inject the error in the ReviewerError like we did at other places (then you can also use the questionmark operator).
Also, passing the result as an argument makes me not so happy. Inserting the error just to get it back out seems odd to me.
Maybe it just would be better to do the database requests from inside the try_review_images method where you can return errors using `?`.

worldofjoni: we should also consider running these image reviews in parallel as 1500 of them could take some (not negligable) amout of time. This can be done with `join_all` over a iterator of futures from the `futures` crate (which is already used in swka_html_request) https://docs.rs/futures/latest/futures/future/fn.join_all.html

backend/src/layer/logic/image_review/image_reviewer.rs
```diff
@@ -0,0 +1,239 @@
...
+        }
+        Ok(())
+    }
+
+    async fn review_image(&self, image: &Image) -> ReviewerResult<()> {
+        let exists = self
+            .image_hoster
+            .check_existence(&image.image_hoster_id)
+            .await
+            .map_err(|_e| ReviewerError::CheckError(image.id))?;
```
worldofjoni, line 102: we dont want to discard the error but inject it in to our ReviewerError. This error is probably more interesting than the image id so you may just make the `CheckError` be created ftom the error of the image hoster using `From`.

backend/src/layer/logic/image_review/mod.rs
```diff
@@ -1 +1,3 @@
 //! This component checks if linked images are still present at the image hoster and removes them from the database if necessary.
+pub mod image_reviewer;
+pub mod test;
```
worldofjoni, line 3:
```suggestion
#[cfg(test)]
mod test;
```

backend/src/layer/logic/image_review/test/image_hoster_mock.rs
```diff
@@ -0,0 +1,54 @@
+use std::sync::{Arc, Mutex};
+
+use async_trait::async_trait;
+
+use crate::interface::image_hoster::{model::ImageMetaData, ImageHoster, ImageHosterError, Result};
+
+pub const PHOTO_ID_TO_FAIL_CHECK_EXISTENCE: &str =
```
worldofjoni, line 7: why do we call these photos now?

backend/src/layer/logic/image_review/test/image_review_database_mock.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl ImageReviewDatabaseMock {
+    #[must_use]
+    pub fn get_n_images_by_rank_date_calls(&self) -> u32 {
+        *self
+            .n_images_by_rank_date_calls
+            .lock()
+            .expect("failed to lock mutex for `n_images_by_rank_date_calls` counter")
+    }
+    #[must_use]
```
worldofjoni, line 30: please add one empty line between functions

backend/src/layer/logic/image_review/test/image_review_database_mock.rs
```diff
@@ -0,0 +1,139 @@
...
+        *self
+            .mark_as_checked_calls
+            .lock()
+            .expect("failed to lock mutex for `mark_as_checked_calls` counter")
+    }
+}
+
+#[async_trait]
+impl ImageReviewDataAccess for ImageReviewDatabaseMock {
+    /// Returns the first n images sorted by rank which are related to an meal served at the given day.
```
worldofjoni, line 56: no doc comments in implementations as they quickly get out of sync.

backend/src/layer/logic/image_review/test/image_review_database_mock.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(images)
+    }
+    /// Returns the first n images sorted by rank which are related to an meal served in the next week or which were not validated last week.
+    async fn get_n_images_next_week_by_rank_not_checked_last_week(
+        &self,
+        n: u32,
+    ) -> Result<Vec<Image>> {
+        *self.n_images_next_week_by_rank_not_checked_last_week_calls.lock().expect("failed to lock mutex for `n_images_next_week_by_rank_not_checked_last_week_calls` counter") += 1;
+
+        let mut images = Vec::new();
```
worldofjoni, line 83: seeing this mutable vec with push and not an iterator hurts 😢 but it is ok in testing code (mock)

backend/src/layer/logic/image_review/test/image_review_database_mock.rs
```diff
@@ -0,0 +1,139 @@
...
+        for _i in 0..n {
```
worldofjoni, line 84: general advice: if you dont use a variable, you can just name it `_` to make clear you discard this information

backend/src/layer/logic/image_review/test/image_review_database_mock.rs
```diff
@@ -0,0 +1,139 @@
...
+            .delete_image_calls
+            .lock()
+            .expect("failed to lock mutex for `delete_image_calls` counter") += 1;
+        if id == ID_TO_THROW_ERROR_ON_DELETE {
+            Err(DataError::NoSuchItem)
+        } else if id == ID_TO_FAIL_DELETE {
+            Ok(false)
+        } else {
+            Ok(true)
+        }
```
worldofjoni, line 128:
```suggestion
        } else  {
            Ok(id != ID_TO_FAIL_DELETE)
        }
```

worldofjoni (changes requested):

backend/src/layer/logic/image_review/image_reviewer.rs
```diff
@@ -0,0 +1,239 @@
...
+        if !exists {
```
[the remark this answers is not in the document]
uglrl: The ordering of the if statements is intentional, because we do not want to delete an image before checking, whether it was deleted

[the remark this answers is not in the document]
uglrl: We could change the signature of the function delete_image, to make it return Result<(), Error>. Then we would not have to handle the case, that the funtion returns false. In which cases would that even occur?

backend/src/layer/logic/image_review/test/image_hoster_mock.rs
uglrl: because of the parameters in check_existence and check_licence. I changed the constants' names. The interfaces need updating

worldofjoni: i dont think trait parameter names matter (to the compiler), so these could just be updated.

backend/src/layer/logic/image_review/image_reviewer.rs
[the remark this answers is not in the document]
worldofjoni: the idea is that false is returned if no image was deleted. But i agree that we can remove this return value.
my ide of reordering was:
```rust
if exists return mark_as_checked;
let deleted = ...;
if !deleted return Err(...);
```

PR #55, opened 2023-07-24
B/mail

uglrl:


backend/src/layer/data/mail/mail_sender.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn get_report(info: &ImageReportInfo) -> MailResult<String> {
+        let template_file_contents =
+            fs::read_to_string(REPORT_TEMPLATE_FILE).map_err(MailError::TemplateError)?;
+        let template = Template::new(&template_file_contents);
+        let mut args = HashMap::new();
+        let image_link: &str = &info.image_link;
+        args.insert("image_link", image_link);
+        let image_id: &str = &Simple::from_uuid(info.image_id).to_string();
```
worldofjoni, line 105: why not just `.to_string()` using hyphonated formatting?
https://docs.rs/uuid/latest/uuid/struct.Uuid.html#formatting

backend/src/layer/data/mail/mail_sender.rs
```diff
@@ -0,0 +1,171 @@
...
+        let image_got_hidden: &str = &info.image_got_hidden.to_string();
+        args.insert("image_got_hidden", image_got_hidden);
+        let positive_rating_count: &str = &info.positive_rating_count.to_string();
+        args.insert("positive_rating_count", positive_rating_count);
+        let negative_rating_count: &str = &info.negative_rating_count.to_string();
+        args.insert("negative_rating_count", negative_rating_count);
+        let get_image_rank: &str = &info.get_image_rank.to_string();
+        args.insert("get_image_rank", get_image_rank);
+
+        Ok(template.render(&args))
```
worldofjoni, line 120: i think it all would be less fo an wall of text if you created a list of pairs and then collected it into a HashMap. All these variable names (which are redundant to the keys in the hashmap) make it realy hard to read.

worldofjoni: maybe even a list of `(&str, &dyn ToString)`

backend/src/startup/config/mail_info.rs
```diff
@@ -0,0 +1,12 @@
+pub struct MailInfo {
```
worldofjoni, line 1: mail info struct should be placed inside mail module (in the data layer)

worldofjoni (changes requested):
looks good for the most part

backend/src/layer/data/mail/mail_sender.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
```
worldofjoni, line 125: consider adding a test to check that the values could be inserted into the template correctrly (and there are no wrongly named fields and so on; on whatever the library errors)

backend/src/layer/data/mail/mail_sender.rs
```diff
@@ -0,0 +1,171 @@
...
+        let sender = self.get_sender()?;
+        let reciever = self.get_reciever()?;
+        let report = Self::get_report(info)?;
+        let email = Message::builder()
+            .from(sender)
+            .to(reciever)
+            .subject("An image was reported for reviewing")
+            .body(report)?;
+        self.mailer.send(&email)?;
+        info!("Email sent successfully!");
```
worldofjoni, line 82: provide more information like an image_id

backend/src/layer/data/mail/mail_sender.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn try_notify_admin_image_report(&self, info: &ImageReportInfo) -> MailResult<()> {
+        let sender = self.get_sender()?;
+        let reciever = self.get_reciever()?;
+        let report = Self::get_report(info)?;
+        let email = Message::builder()
+            .from(sender)
+            .to(reciever)
+            .subject("An image was reported for reviewing")
```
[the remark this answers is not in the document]
uglrl: An image was reported and requires your review

backend/src/layer/data/mail/mail_sender.rs
```diff
@@ -0,0 +1,183 @@
...
+    const SMTP_PORT_ENV_NAME: &str = "SMTP_PORT";
+    const SMTP_USERNAME_ENV_NAME: &str = "SMTP_USERNAME";
+    const SMTP_PASSWORD_ENV_NAME: &str = "SMTP_PASSWORD";
+    const ADMIN_EMAIL_ENV_NAME: &str = "ADMIN_EMAIL";
+
+    #[tokio::test]
+    async fn test_get_report() {
+        let info = get_report_info();
+        let report = MailSender::get_report(&info).replace("\r\n", "\n");
+        let expected = format!("The image at the url {}\nwith the id {}\nwas reported {} times.\nReason: {}\nImage automatically hidden: {}\n\nAdditional Data:\nPositive ratings: {}\nNegative ratings: {}\nRank: {}",
```
worldofjoni, line 135: that was not what i meant with testing the formating.
If there is no way to get an error from the template render method, you could check whether any pattern string using `{{ ...}}` (or whatever its syntax was) sill remains in the format (maybe if you missspelled an argument in the template file).
We dont want to change the test if we change the template.

worldofjoni (approved):

PR #59, opened 2023-07-24
B/database

worldofjoni:

## Backend

## TODO
- [x] Refactor `as`, nullable Options and similar
- [x] Tests, tests, tests (they are very important here)


backend/src/layer/data/database/request.rs
```diff
@@ -0,0 +1,451 @@
...
+            return Ok(None);
+        }
+
+        let meal = sqlx::query!(
+            r#"
+            SELECT food_id, name, food_type as "meal_type: MealType",
+                price_student, price_employee, price_guest, price_pupil, serve_date as date, line_id,
+                new, frequency, last_served, next_served, average_rating, rating_count
+            FROM meal_detail JOIN food_plan USING (food_id)
+            WHERE line_id = $1 AND serve_date = $2
```
uglrl, line 119: Consider adding some ordering here. This makes the tests easier and the frontend more consistent

backend/README.md
```diff
@@ -36,9 +36,13 @@ For deployment of the server, initial settings like API tokens and other access
...
+1. Setup your environment variable for the database. The default should be ok when you use the command below for your database.
+
+1. Install docker and run to spun up a database:
+    ```bash
+    docker run -itd -e POSTGRES_USER=postgres_user -e POSTGRES_PASSWORD=secret_password -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_DB=mensa_app -p 5432:5432 -v data:/var/lib/postgresql/data --name postgresql postgres
+    ```
+    This runs a postgres database as a docker container.
+    To setup all relations install `cargo install sqlx-cli` and run `cargo sqlx mig run`.
+
+1. After adding a query in rust code, make sure to run `cargo sqlx prepare` to safe the database format for offline usage so compilation can be successful even if no database is present.
```
Whatsuup, line 48: add `cargo sqlx database reset` as it seems useful for debugging

backend/migrations/20230703125429_canteen.up.sql
```diff
@@ -1,5 +1,6 @@
 -- Add up migration script here
 CREATE TABLE canteen (
   canteen_id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
-  name text NOT NULL
+  name text NOT NULL,
+  position integer NOT NULL
```
Whatsuup, line 5: It seems practical to define a default position.

backend/migrations/20230703130438_food.up.sql
```diff
@@ -2,7 +2,7 @@
 CREATE TYPE meal_type AS ENUM ('VEGAN','VEGETARIAN','BEEF','BEEF_AW','PORK','PORK_AW','FISH','UNKNOWN');
 
 CREATE TABLE food (
-  food_id uuid PRIMARY KEY,
+  food_id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
   name text NOT NULL,
   food_type meal_type NOT NULL
```
Whatsuup, line 7: Default meal_type unnkown?

backend/migrations/20230724125220_trigram.up.sql
```diff
@@ -0,0 +1,4 @@
+-- Add up migration script here
+
+CREATE EXTENSION pg_trgm;
+SET pg_trgm.similarity_threshold = 0.8;
```
Whatsuup, line 4: Remove as there's no longer a need for? If we don't get it to run, we should remove it.

backend/migrations/20230724125220_trigram.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Add down migration script here
+DROP EXTENSION pg_trgm;
```
Whatsuup, line 2: Remove as there's no longer a need for? If we don't get it to run, we should remove it.

backend/src/interface/persistent_data/model.rs
```diff
@@ -31,6 +34,7 @@ pub struct Line {
 }
 
 /// Struct for database-operations. Related to the database entity 'meal'.
+#[derive(Debug, PartialEq)]
```
Whatsuup, line 37: add Eq aswell

backend/src/interface/persistent_data/model.rs
```diff
@@ -70,7 +75,7 @@ pub struct Side {
     pub price: Price,
 }
 
-#[derive(Default)]
+#[derive(Debug, Default, PartialEq, Clone)]
```
Whatsuup, line 78: add Eq aswell

backend/src/interface/persistent_data.rs
```diff
@@ -17,96 +15,139 @@ pub enum DataError {
...
         name: &str,
         meal_type: MealType,
-        price: Price,
-        next_served: Date,
         allergens: &[Allergen],
         additives: &[Additive],
-    ) -> Result<Side>;
+    ) -> Result<Uuid>;
+
+    /// Adds a meal into the meal plan with the given params.
```
uglrl, line 121: Bad documentation

uglrl (changes requested):

backend/src/interface/persistent_data/model.rs
uglrl: That won't work, because meal contains an f32, which cannot be Eq'ed

backend/src/interface/persistent_data/model.rs
uglrl: Same problem as with Meal: f32 cannot be Eq'ed

backend/src/interface/persistent_data.rs
```diff
@@ -17,96 +15,139 @@ pub enum DataError {
...
+
     /// Determines the canteen with the most similar name.
-    async fn get_similar_canteen(&self, similar_name: &str) -> Result<Option<Canteen>>;
+    /// Returns the UUID to the similar canteen.
+    async fn get_similar_canteen(&self, similar_name: &str) -> Result<Option<Uuid>>;
+
     /// Determines the line with the most similar name.
-    async fn get_similar_line(&self, similar_name: &str) -> Result<Option<Line>>;
+    /// Returns the UUID to the similar line.
+    async fn get_similar_line(&self, similar_name: &str) -> Result<Option<Uuid>>;
```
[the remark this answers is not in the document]
Whatsuup: There is no more information needed for the further procedure. Aside: It is better to reduce data as return too much and a lot of information keeps untouched.

backend/src/layer/data/database/command.rs
```diff
@@ -0,0 +1,420 @@
...
+        Ok(Image {
+            approved: null_error!(record.approved),
+            url: null_error!(record.url),
+            rank: null_error!(record.rank),
+            report_count: u32::try_from(null_error!(record.report_count))?,
+            upload_date: null_error!(record.upload_date),
+            downvotes: u32::try_from(null_error!(record.downvotes))?,
+            upvotes: u32::try_from(null_error!(record.upvotes))?,
+            id: null_error!(record.image_id),
+            image_hoster_id: null_error!(record.image_hoster_id),
```
Whatsuup, line 44: These null checker only appear here, why?

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -0,0 +1,842 @@
...
+use crate::{
+    interface::persistent_data::{MealplanManagementDataAccess, Result},
+    util::{Additive, Allergen, Date, MealType, Price, Uuid},
+};
+
+pub struct PersistentMealplanManagementData {
+    pub(super) pool: Pool<Postgres>,
+}
+
+const THRESHOLD: f32 = 0.785;
```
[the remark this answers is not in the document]
Whatsuup: Can be done and seems right. All tests work with the same threshold. If you got some cases that would not fit with that limit, we can change that.

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -0,0 +1,842 @@
...
+            ",
+            date,
+            canteen_id
+        )
+        .execute(&self.pool)
+        .await?;
+        Ok(())
+    }
+
+    async fn get_similar_canteen(&self, similar_name: &str) -> Result<Option<Uuid>> {
```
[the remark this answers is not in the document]
Whatsuup: First, it is impossible that there are 3 similar meals, because this function exists. But if there are three equal canteens, the database decides which canteen will be in the first row. And as there is a system beneath, the canteen will always be the same.

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -0,0 +1,842 @@
...
+        .fetch_one(&self.pool)
+        .await
+        .map_err(Into::into)
+    }
+
+    async fn update_line(&self, uuid: Uuid, name: &str, position: u32) -> Result<Uuid> {
+        sqlx::query_scalar!(
+            "
+            UPDATE line
+            SET name = $2, position = $3
```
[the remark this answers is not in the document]
Whatsuup: Canteens will be ordered by the env string. Lines could be a problem, if a line changes its canteen. We could add to similarity test, that the canteen_id must be similar. So each canteen has its own set of lines.
Good point.

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -0,0 +1,842 @@
...
+        name: &str,
+        meal_type: MealType,
+        allergens: &[Allergen],
+        additives: &[Additive],
+    ) -> Result<Uuid> {
+        self.insert_food(name, meal_type, allergens, additives)
+            .await
+    }
+
+    async fn add_meal_to_plan(
```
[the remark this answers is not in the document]
Whatsuup: This is only here the case. All above have small changes. I suggest keeping it like that, as it is easier to change them.

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -0,0 +1,842 @@
...
+                Uuid::parse_str("10728cc4-1e07-4e18-a9d9-ca45b9782413").unwrap(),
+                "chicco di caffé Karlsruhe",
+                true,
+            ),
+            (
+                Uuid::parse_str("f2885f67-fc95-4205-bc7d-b2fb78cee0a8").unwrap(),
+                "Cafebar Moltke 2",
+                true,
+            ),
+            // No longer 'similar'
```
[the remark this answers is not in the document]
Whatsuup: The strings change a little bit from top to down

backend/src/layer/data/database/request.rs
```diff
@@ -0,0 +1,649 @@
...
+
+    async fn get_sides(&self, line_id: Uuid, date: Date) -> Result<Vec<Side>> {
+        sqlx::query!(
+            r#"
+            SELECT food_id, name, food_type as "meal_type: MealType", 
+            price_student, price_employee, price_guest, price_pupil
+            FROM food JOIN food_plan USING (food_id)
+            WHERE line_id = $1 AND serve_date = $2 AND food_id NOT IN (SELECT food_id FROM meal)
+            ORDER BY food_id
+            "#,
```
Whatsuup, line 162: order by uuid is random

Whatsuup (changes requested):
The two "breaking" changes detected by peer must be solved!

backend/src/layer/data/database/request.rs
uglrl: The ordering is (mostly) for testing purposes. The order was random before, now we have something to compare

backend/migrations/20230703125429_canteen.up.sql
uglrl: which position would you choose for that? 0 or a very large number?

backend/src/interface/persistent_data.rs
```diff
@@ -17,96 +15,139 @@ pub enum DataError {
...
     async fn get_n_images_by_rank_date(&self, n: u32, date: Date) -> Result<Vec<Image>>;
-    /// Returns the first n images sorted by rank which are related to an meal served in the next week or which were not validated last week.
+    /// Returns the first n images sorted by rank which are related to an meal served in the next week and which were not validated last week.
     async fn get_n_images_next_week_by_rank_not_checked_last_week(
         &self,
         n: u32,
     ) -> Result<Vec<Image>>;
-    /// Returns the first n images sorted by the date of the last check (desc) which were not validated in the last week.
+    /// Returns the first n images sorted by the date of the last validation (asc) which were not validated in the last week.
     async fn get_n_images_by_last_checked_not_checked_last_week(
```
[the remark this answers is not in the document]
worldofjoni: what would be a better name? i dont think something like `get_review_images_part_1` would be better

backend/migrations/20230724125220_trigram.down.sql
worldofjoni: ?

backend/src/interface/persistent_data.rs
worldofjoni: what would you wish?

backend/src/layer/data/database/request.rs
worldofjoni: this ordering is exactly the opposite of random. It is deterministic but a user may not nee a pattern. This is just to avoid foods randomly switching order when the db desides to. performance whise sorting by an index should be ok (add no significant overhead)

backend/src/layer/data/database/mealplan_management.rs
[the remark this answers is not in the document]
worldofjoni: we cloud add a UNIQUE constrain to (canteen_id, position) in the database. But this is problematic as lines may change and then a new line could not be inserted (which is a case we then have to handle; how?). I think it is better thath when a line is replaced the app shows both lines at the same position instead of not showing the new line at all.

backend/migrations/20230703125429_canteen.up.sql
worldofjoni: if we want position to be unique, we cannot /should not do that. even if we dont want it to be enforced to be unique, i dont think a default position makes sense.

backend/src/interface/persistent_data.rs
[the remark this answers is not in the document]
uglrl: get_images_not_checked_last_week_sorted_by_validation_date(number_of_images)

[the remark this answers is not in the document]
uglrl: Wofür brauchen wir hier 3 Funktionen?

backend/src/layer/data/database/image_review.rs
```diff
@@ -0,0 +1,315 @@
...
+                approved: null_error!(r.approved),
+                report_count: u32::try_from(null_error!(r.report_count))?,
+                upload_date: null_error!(r.link_date),
+            })
+        })
+        .collect::<Result<Vec<_>>>()
+    }
+
+    async fn delete_image(&self, id: Uuid) -> Result<()> {
+        // Todo on delete cascade?
```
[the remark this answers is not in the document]
uglrl: Ich glaube der test läuft trotzdem durch, da wir keine reports oder bewertungen included hatten

backend/src/interface/persistent_data.rs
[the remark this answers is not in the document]
uglrl: get_images_for_date
get_unvalidated_images_for_next_week
get_old_images

backend/src/interface/persistent_data.rs
```diff
@@ -17,96 +15,139 @@ pub enum DataError {
...
+    /// Updates an existing canteen entity in the database.
+    /// Returns the canteen's UUID.
+    async fn update_canteen(&self, uuid: Uuid, name: &str, position: u32) -> Result<Uuid>;
+
+    /// Updates an existing line entity in the database.
+    /// Returns the line UUID.
+    async fn update_line(&self, uuid: Uuid, name: &str, position: u32) -> Result<Uuid>;
+
+    /// Updates an existing meal entity in the database.
+    async fn update_meal(&self, uuid: Uuid, name: &str) -> Result<()>;
```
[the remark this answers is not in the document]
uglrl: @Whatsuup

backend/src/layer/data/database/mealplan_management.rs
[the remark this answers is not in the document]
uglrl: @Whatsuup

Whatsuup (approved):

PR #73, opened 2023-08-03
B/better loging

worldofjoni:

### DIFF
--- a/backend/src/layer/data/flickr_api/api_request.rs
+++ b/backend/src/layer/data/flickr_api/api_request.rs
@@ -3,7 +3,7 @@ use crate::interface::image_hoster::ImageHosterError;
 use crate::layer::data::flickr_api::json_parser::JsonParser;
 use crate::layer::data::flickr_api::json_structs::{JsonRootError, JsonRootLicense, JsonRootSizes};
 use reqwest::Response;
-use tracing::debug;
+use tracing::trace;
 
 pub struct ApiRequest {
     api_key: String,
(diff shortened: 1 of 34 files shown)


backend/src/layer/data/flickr_api/api_request.rs
Whatsuup, line 1: These functions changed. Changes will be resolved during merge.

backend/src/layer/data/flickr_api/api_request.rs
```diff
@@ -88,11 +88,10 @@ impl ApiRequest {
     }
```
Whatsuup, line 89: Google won't give me such short and easy solutions...

Whatsuup (commented):
I expect these graphql request traces cause too much log traffic. Should we still log this information in another mode?

worldofjoni:
@Whatsuup 
Trace is already the lowes mode.
We now only trace every request once. I think this is ok, because if you dont want to get all these messages, you can just disable the trace level.

backend/src/layer/trigger/graphql/server.rs
```diff
@@ -164,14 +167,42 @@ async fn graphql_handler(
...
     request: GraphQLRequest,
 ) -> GraphQLResponse {
-    let request = request.into_inner().data(
-        headers
-            .get("Authorization")
-            .and_then(|h| h.to_str().ok())
-            .unwrap_or("")
-            .to_string() as AuthHeader,
+    let auth_header = headers
+        .get("Authorization")
```
worldofjoni, line 171: constant?

Whatsuup (approved):

PR #75, opened 2023-08-03
B/unit/alex

Whatsuup:
## Backend
- [ ] You added **unit tests** to all your implemented functions.
- [ ] You ran `cargo format` to format your code.

### DIFF
--- a/backend/src/layer/data/flickr_api/flickr_api_handler.rs
+++ b/backend/src/layer/data/flickr_api/flickr_api_handler.rs
@@ -26,7 +26,7 @@ lazy_static! {
 
 impl FlickrApiHandler {
     #[must_use]
-    #[allow(clippy::missing_const_for_fn)] // cannot be made const because of compiler error`
+    #[allow(clippy::missing_const_for_fn)]
     pub fn new(info: FlickrInfo) -> Self {
         Self {
             request: ApiRequest::new(info.api_key),
@@ -121,13 +121,21 @@ where
 #[cfg(test)]
 mod test {
     #![allow(clippy::unwrap_used)]
+
+    use std::env;
     use std::sync::Arc;
 
     use async_trait::async_trait;
+    use dotenvy::dotenv;
 
     use crate::interface::image_hoster::model::ImageMetaData;
     use crate::interface::image_hoster::{ImageHoster, ImageHosterError, Result};
-    use crate::layer::data::flickr_api::flickr_api_handler::FlickrApiHandler;
+    use crate::layer::data::flickr_api::flickr_api_handler::{FlickrApiHandler, FlickrInfo};
+
+    fn get_api_key() -> String {
+        dotenv().ok();
+        env::var("FLICKR_API_KEY").expect("FLICKR_API_KEY should be set in the .env!")
+    }
 
     #[test]
     fn test_valid_determine_short_photo_id() {
@@ -137,29 +145,29 @@ mod test {
     }
 
     #[test]
-    fn test_valid_determine_long_photo_id() {
-        let valid_url = "https://www.flickr.com/photos/198319418@N06/53077317043/";
+    fn test_valid_determine_long1_photo_id() {
+        let valid_url = "https://www.flickr.com/photos/198319418@N06/53077317043";
         let res = FlickrApiHandler::determine_photo_id(valid_url).unwrap();
         assert_eq!(res, "53077317043");
     }
 
     #[test]
-    fn test_empty_determine_photo_id() {
-        let valid_url = "";
-        let res = FlickrApiHandler::determine_photo_id(valid_url)
-            .err()
-            .unwrap();
-        assert_eq!(
-            res,
-            ImageHosterError::FormatNotFound(format!(
-                "this url format is not supported: '{valid_url}'"
-            ))
-        );
+    fn test_valid_determine_long2_photo_id() {
+        let valid_url = "https://www.flickr.com/photos/gerdavs/52310534489/";
+        let res = FlickrApiHandler::determine_photo_id(valid_url).unwrap();
+        assert_eq!(res, "52310534489");
     }
 
     #[test]
     fn test_invalid_determine_photo_id() {
-        let valid_url = "https://flic.kr/p/";
+        let valid_url = "https://www.flickr.com/photos/gerdavs/";
+        let res = FlickrApiHandler::determine_photo_id(valid_url);
+        assert!(res.is_err());
+    }
+
+    #[test]
+    fn test_empty_determine_photo_id() {
+        let valid_url = "";
         let res = FlickrApiHandler::determine_photo_id(valid_url)
             .err()
             .unwrap();
@@ -197,5 +205,30 @@ mod test {
         arc.check_existence("image_id").await.unwrap();
         arc.check_licence("image_id").await.unwrap();
         arc.validate_url("url").await.unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_validate_url() {
+        let handler = FlickrApiHandler::new(FlickrInfo {
+            api_key: get_api_key(),
+        });
+        let url = "https://www.flickr.com/photos/198319418@N06/53077317043";
+        assert!(handler.validate_url(url).await.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_check_license() {
+        let handler = FlickrApiHandler::new(FlickrInfo {
+            api_key: get_api_key(),
+        });
+        assert!(handler.check_licence("53077317043").await.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_check_existence() {
+        let handler = FlickrApiHandler::new(FlickrInfo {
+            api_key: get_api_key(),
+        });
+        assert!(handler.check_existence("53077317043").await.is_ok());
     }
 }
--- a/backend/src/layer/data/flickr_api/flickr_api_handler.rs
+++ b/backend/src/layer/data/flickr_api/flickr_api_handler.rs
@@ -18,7 +18,7 @@ pub struct FlickrApiHandler {
 
 lazy_static! {
     static ref LONG_URL_REGEX: Regex =
-        Regex::new(r"(https://www.flickr.com/photos/)(\w+)/(\d+)([/]{0,1})")
+        Regex::new(r"(https://www.flickr.com/photos/)([\d\w@]+)/(\d+)([/]{0,1})")
             .expect("regex creation failed");
     static ref SHORT_URL_REGEX: Regex =
         Regex::new(r"(https://flic.kr/p/)([\d\w]+)").expect("regex creation failed");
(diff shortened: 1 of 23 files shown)


backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,22 +18,23 @@ pub struct FlickrApiHandler {
 
 lazy_static! {
     static ref LONG_URL_REGEX: Regex =
-        Regex::new(r"(https://www.flickr.com/photos/)(\w+)/(\d+)([/]{0,1})")
+        Regex::new(r"(https://www.flickr.com/photos/)([\d\w@]+)/(\d+)([/]{0,1})")
```
worldofjoni, line 21: regex `?` for 0 or 1
you also only need to put paranteses "()" around blocks if you want to group then to extract their content later

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,22 +18,23 @@ pub struct FlickrApiHandler {
...
             .expect("regex creation failed");
     static ref SHORT_URL_REGEX: Regex =
         Regex::new(r"(https://flic.kr/p/)([\d\w]+)").expect("regex creation failed");
 }
 
 impl FlickrApiHandler {
     #[must_use]
-    #[allow(clippy::missing_const_for_fn)] // cannot be made const because of compiler error`
+    #[allow(clippy::missing_const_for_fn)]
```
worldofjoni, line 29: i think a comment there would be helpful

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,22 +18,23 @@ pub struct FlickrApiHandler {
...
             request: ApiRequest::new(info.api_key),
         }
     }
 
-    // URL TYPE 1: https://www.flickr.com/photos/gerdavs/52310534489/ <- remove last '/'
+    // URL TYPE 1.1: https://www.flickr.com/photos/gerdavs/52310534489/ <- remove last '/'
+    // URL TYPE 1.2: https://www.flickr.com/photos/198319418@N06/53077317043 <- remove last '/'
     // URL TYPE 2: https://flic.kr/p/2oRguN3
     // Both cases: Split with '/' and get last member (= photo_id).
     fn determine_photo_id(mut url: &str) -> Result<&str> {
```
worldofjoni, line 40: why did you go away from using regex capute roups to get the id?
This would be more reliable and easier to adapt

backend/src/layer/data/swka_parser/swka_link_creator.rs
```diff
@@ -132,4 +134,12 @@ mod tests {
         let result = test_util::get_creator().get_all_urls_for_next_weeks_from_date(date);
         assert_eq!(result, URLS_FOR_NEXT_WEEKS);
     }
+
+    #[test]
+    fn test_get_todays_date() {
```
worldofjoni, line 139: i dnt think this is a helpful test (testing a function against its implementation). If you want to get coverage for this function, there should be a test which tests the function where this is called

worldofjoni (changes requested):

backend/src/layer/data/flickr_api/flickr_api_handler.rs
Whatsuup: i want to reduce regex checks in the function (like `if matches1 || matches2 ...`)

backend/src/layer/data/flickr_api/flickr_api_handler.rs
Whatsuup: https://docs.rs/regex/latest/regex/struct.Captures.html
That is something new to me... but very helpful. This will be implemented in the next push.

backend/src/layer/data/swka_parser/swka_link_creator.rs
Whatsuup: It is for sure not. The problem is, we can only test this function with date related tests. So test only will work for one day. There's no "easy" way to create a test case that calls this function. We can remove it, as it is very dumb, but good for coverage :)

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,35 +18,66 @@ pub struct FlickrApiHandler {
...
+                Self::decode(&id)
+            } else {
+                Ok(id)
+            }
+        } else {
+            Err(ImageHosterError::FormatNotFound(format!(
                 "this url format is not supported: '{url}'"
-            )));
+            )))
+        }
```
Whatsuup, line 63: Hear me out:
bs58 provides decoding functionality with the FLICKR alphabet. BUT the library (as far as I know) only parses strings into vec<u8> (utf8). Flickr wants to calculate an integer value that depends on indexes. As Bs58 is not capable of it and the implementation is not that hard, I suggest to do it by our selfs.
(See https://www.dcode.fr/base-58-cipher {short id} -> into Integer)

backend/src/layer/data/swka_parser/swka_link_creator.rs
worldofjoni: is there no function calling `get_all_urls`?
if you only call this function once (even if not) you could imo just inline it.

backend/src/layer/data/flickr_api/flickr_api_handler.rs
worldofjoni: encodings like base58, base64 etc are ment to encode binary data as strings (and in reverse).
Thus you need to _decode_ the short url and get binary data back. This binary data you then have to interpret as a number (as flickr ids are numbers) and convert it into a string for our representation.

worldofjoni: for this conversion you have to figure out the byte ordering (Big endien or little endien) used

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -37,17 +39,45 @@ impl FlickrApiHandler {
...
+                .ok_or_else(|| {
+                    ImageHosterError::FormatNotFound(format!(
+                        "this url format is not supported: '{url}'"
+                    ))
+                })?;
+            if SHORT_URL_REGEX.is_match(url) {
+                Self::decode(&id)
+            } else {
+                Ok(id)
+            }
```
worldofjoni, line 58: you unnseccsearily call regex.is_match two times for SHORT_URL_REGEX.
You can just split these into two seperate cases.
The you also could easily let the regex do the extracting for you and you dont have to do the whole tim_end, split part.

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -37,17 +39,45 @@ impl FlickrApiHandler {
...
-        })
+    }
+
+    fn decode(word: &str) -> Result<String> {
+        let mut decoded = 0_u64;
+        let mut multi = 1_u64;
+        for char in word.chars().rev() {
+            let index = ALPHABET.find(char).ok_or_else(|| {
+                ImageHosterError::FormatNotFound(String::from(
+                    "FlickrDecoder: Provided photo_id contains invalid characters.",
```
worldofjoni, line 72: general advice: an good error message would at least contain the invalid char

worldofjoni (changes requested):

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,35 +18,74 @@ pub struct FlickrApiHandler {
...
+    // URL TYPE 2: https://flic.kr/p/2oRguN3, id group = 2
     // Both cases: Split with '/' and get last member (= photo_id).
-    fn determine_photo_id(mut url: &str) -> Result<&str> {
-        if !SHORT_URL_REGEX.is_match(url) && !LONG_URL_REGEX.is_match(url) {
-            return Err(ImageHosterError::FormatNotFound(format!(
+    fn determine_photo_id(url: &str) -> Result<String> {
+        if SHORT_URL_REGEX.is_match(url) {
+            let groups = SHORT_URL_REGEX
+                .captures(url)
+                .expect("url matches regex but captures could not be created");
```
Whatsuup, line 46: I could not imagine a scenario where this captures creation can fail

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,35 +18,74 @@ pub struct FlickrApiHandler {
...
+                        "could not detect id group in '{url}'"
+                    )))
+                },
+                Self::decode,
+            )
+        } else if LONG_URL_REGEX.is_match(url) {
+            let groups = LONG_URL_REGEX
+                .captures(url)
+                .expect("url matches regex but captures could not be created");
+            groups.get(4).map(|m| m.as_str()).map_or_else(
```
Whatsuup, line 59: How to handle these 'magic-numbers'?

backend/src/layer/data/flickr_api/flickr_api_handler.rs
worldofjoni: you could (and should) simplify the if expression: the `captures` method returns none if the regex does not match.
`if let`would be a good fit here.

backend/src/layer/data/flickr_api/flickr_api_handler.rs
worldofjoni: as you only need to extract one bit out of the string, you could use avoid using `()` everywhere else. Then you would only have one capture group, you can index with 1 (0 is always the entire match string). If you then can guarantee, that such an capture group always exists (which you can with the regex provided), you can simply access it by `groups[1]` and dont have to work with an option.

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,66 +18,60 @@ pub struct FlickrApiHandler {
...
-                    ImageHosterError::FormatNotFound(format!(
-                        "this url format is not supported: '{url}'"
-                    ))
-                })?;
-            if SHORT_URL_REGEX.is_match(url) {
-                Self::decode(&id)
-            } else {
-                Ok(id)
+        if let Some(groups) = SHORT_URL_REGEX.captures(url) {
+            match groups.get(1).map(|m| m.as_str()) {
```
worldofjoni, line 42: get rid of none case with expect, as this can never happen with our regex

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -18,66 +18,60 @@ pub struct FlickrApiHandler {
...
-                multi * u64::try_from(index).expect("FlickrDecoder: Could not parse usize to u64");
-            multi *=
-                u64::try_from(ALPHABET.len()).expect("FlickrDecoder: Could not parse usize to u64");
-        }
-        Ok(decoded.to_string())
+    fn decode(word: &str) -> Result<u64> {
+        let bytes = bs58::decode(word)
+            .with_alphabet(bs58::Alphabet::FLICKR)
+            .into_vec()
+            .map_err(|e| ImageHosterError::DecodeFailed(e.to_string()))?;
```
worldofjoni, line 68: seperate error for this case, something about invalid short url

worldofjoni (changes requested):

worldofjoni (approved):

PR #80, opened 2023-08-09
B/fixes

Whatsuup:


backend/migrations/20230703130648_food_additive.up.sql
```diff
@@ -1,5 +1,5 @@
 -- Add up migration script here
-CREATE TYPE additive AS ENUM ('COLORANT','PRESERVING_AGENTS','ANTIOXIDANT_AGENTS','FLAVOUR_ENHANCER','PHOSPHATE','SURFACE_WAXED','SULPHUR','ARTIFICALLY_BLACKENED_OLIVES','SWEETENER','LAXATIVE_IF_OVERUSED','PHENYLALANINE','ALCOHOL','PRESSED_MEAT','GLAZING_WITH_CACAO','PRESSED_FISH');
+CREATE TYPE additive AS ENUM ('COLORANT','PRESERVING_AGENTS','ANTIOXIDANT_AGENTS','FLAVOUR_ENHANCER','PHOSPHATE','SURFACE_WAXED','SULPHUR','ARTIFICIALLY_BLACKENED_OLIVES','SWEETENER','LAXATIVE_IF_OVERUSED','PHENYLALANINE','ALCOHOL','PRESSED_MEAT','GLAZING_WITH_CACAO','PRESSED_FISH');
```
worldofjoni, line 2: this rename will be difficult to migrate withut loosing all data in the database...

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -85,6 +86,7 @@ impl ImageHoster for FlickrApiHandler {
     /// # Return
     /// If the image exists, the [`ImageMetaData`] struct will be returned.
     async fn validate_url(&self, url: &str) -> Result<ImageMetaData> {
+        trace!("FlickrApi: Started url validation with '{url}'");
```
worldofjoni, line 89: too many log messages!
i would remove them here and keep those when requestig external sources

worldofjoni (changes requested):

backend/src/layer/logic/api_command/command_handler.rs
```diff
@@ -169,19 +170,34 @@ where
...
+                    Err(e) => {
+                        warn!("{e}");
+                        return Err(CommandError::ImageHosterError(e));
+                    }
+                }
+            }
+            Err(e) => {
+                warn!("{e}");
+                return Err(CommandError::ImageHosterError(e));
+            }
```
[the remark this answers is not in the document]
worldofjoni: like i said, i think its best to change the signature of `check_license` to `-> Result<()>` and simply return an error when the license is invalid. In this Error you then can put a pretty message with the given and available livenses

backend/migrations/20230703130648_food_additive.up.sql
worldofjoni: ok, renaming enum variants is possible. Should we do that as a migration or change thah here and run the rename in the production db manually?

Whatsuup: manually change if possible

backend/src/interface/image_hoster.rs
```diff
@@ -24,6 +24,9 @@ pub enum ImageHosterError {
     /// Photo not found error
     #[error("the photo id passed was not a valid photo id")]
     PhotoNotFound,
+    /// License invalid
+    #[error("the detected license cannot be used: {0}")]
+    InvalidLicense(String),
```
worldofjoni, line 29: also add list of valid licenses

backend/src/layer/data/flickr_api/json_parser.rs
```diff
@@ -55,17 +55,17 @@ impl JsonParser {
     /// A boolean if the image has an valid license or not.
     /// If the image has no license or no license history, the image isn't restricted by any license and true will be returned.
     #[must_use]
-    pub fn check_license(root: &JsonRootLicense) -> bool {
+    pub fn check_license(root: &JsonRootLicense) -> (bool, String) {
```
worldofjoni, line 58: a more idiomatic return type would be `Result<(), String>` or ever better produce the error here.

backend/src/layer/data/flickr_api/json_parser.rs
```diff
@@ -55,17 +55,17 @@ impl JsonParser {
...
         let license = root
             .license_history
             .iter()
             .max_by_key(|l| l.date_change)
             .map(|entry| &entry.new_license);
 
         if let Some(license) = license {
-            return VALID_LICENSES.contains(&license.as_str());
+            return (VALID_LICENSES.contains(&license.as_str()), license.clone());
```
worldofjoni, line 66: no need to clone.
if the &license for .contains makes problems, you could just put that into an extra line before.

worldofjoni (changes requested):

backend/src/interface/image_hoster.rs
```diff
@@ -24,7 +24,10 @@ pub enum ImageHosterError {
     /// Photo not found error
     #[error("the photo id passed was not a valid photo id")]
     PhotoNotFound,
-    /// License invalid
+    /// License invalid. List of valid licenses:
+    /// - `No known copyright restrictions`
+    /// - `Public Domain Dedication (CC0)`
+    /// - `Public Domain Mark`
```
worldofjoni, line 30: a thought of putting these *inside* the error message as this would be helpful for api users and be not redundant when we just use the variable.

backend/src/layer/data/flickr_api/flickr_api_handler.rs
```diff
@@ -118,12 +118,7 @@ impl ImageHoster for FlickrApiHandler {
...
     async fn check_licence(&self, photo_id: &str) -> Result<()> {
-        let (valid, license) = self.request.flickr_photos_license_check(photo_id).await?;
-        return if valid {
-            Ok(())
-        } else {
-            Err(ImageHosterError::InvalidLicense(license))
-        };
+        self.request.flickr_photos_license_check(photo_id).await
     }
 }
```
worldofjoni, line 123: cant this method be just removed now?

backend/src/layer/data/flickr_api/json_parser.rs
```diff
@@ -54,18 +54,26 @@ impl JsonParser {
...
+            return if VALID_LICENSES.contains(&str_license) {
+                Ok(())
+            } else {
+                Err(ImageHosterError::InvalidLicense(license))
+            };
         }
-        (false, String::from("No license could be detected"))
+        Err(ImageHosterError::InvalidLicense(String::from(
+            "No license could be detected",
+        )))
```
worldofjoni, line 76: reorder: put option with `.ok_or(Error...)?` first and return it.
then no `return` is needed as it all can just be the last line.

worldofjoni (changes requested):

backend/src/layer/data/flickr_api/flickr_api_handler.rs
Whatsuup: No?

worldofjoni (changes requested):

backend/src/layer/data/flickr_api/json_parser.rs
```diff
@@ -54,18 +54,29 @@ impl JsonParser {
...
+                )
+            })?;
+        let str_license = license.as_str();
+        if VALID_LICENSES.contains(&str_license) {
+            Ok(())
+        } else {
+            Err(ImageHosterError::InvalidLicense(
+                license,
+                format!("{VALID_LICENSES:?}"),
+            ))
```
Whatsuup, line 78: Is there a way to pass these arrays without debug print? Or is the only aesthetic solution to implement display?

worldofjoni: an array of strings can be `.join(delimiter)`ed to a single string

worldofjoni:
are these uncoverd lines something we can/should test

worldofjoni (approved):

PR #87, opened 2023-08-18
B/fixes

Whatsuup:


backend/src/layer/data/flickr_api/json_parser.rs
```diff
@@ -182,16 +183,49 @@ mod test {
...
+                },
+                LicenceHistory {
+                    date_change: 1_598_990_519,
+                    new_license: String::from("All Rights Reserved"),
+                    old_license: String::from("United States Government Work"),
+                },
+            ],
+        };
+        assert!(JsonParser::check_license(valid_licenses).is_err());
+    }
```
Whatsuup, line 227: Short info:
As you can see, only the last history will be checked. 
So only the last old_license and the last new_license is used for validation

worldofjoni (approved):

PR #90, opened 2023-08-20
B/int/image scheduler

uglrl:

### DIFF
--- a/backend/tests/image_review_scheduler.rs
+++ b/backend/tests/image_review_scheduler.rs
@@ -1,39 +1,51 @@
+use async_trait::async_trait;
 use mensa_app_backend::{
+    interface::mealplan_management::MensaParseScheduling,
     layer::{
         data::{
             database::factory::DataAccessFactory, flickr_api::flickr_api_handler::FlickrApiHandler,
-            swka_parser::swka_parse_manager::SwKaParseManager,
         },
-        logic::{
-            image_review::image_reviewer::ImageReviewer,
-            mealplan_management::meal_plan_manager::MealPlanManager,
-        },
+        logic::image_review::image_reviewer::ImageReviewer,
         trigger::scheduling::scheduler::{ScheduleInfo, Scheduler},
     },
     startup::config::ConfigReader,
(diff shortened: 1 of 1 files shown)


backend/tests/image_review_scheduler.rs
```diff
@@ -0,0 +1,54 @@
...
+            database::factory::DataAccessFactory, flickr_api::flickr_api_handler::FlickrApiHandler,
+        },
+        logic::image_review::image_reviewer::ImageReviewer,
+        trigger::scheduling::scheduler::{ScheduleInfo, Scheduler},
+    },
+    startup::config::ConfigReader,
+};
+use tokio::signal::ctrl_c;
+
+const NEVER: &str = "* * * 31 2 *";
```
worldofjoni, line 15: xD

backend/tests/image_review_scheduler.rs
```diff
@@ -0,0 +1,54 @@
...
+        image_review_schedule: "0 */5 * * * *".to_string(),
+    };
+    let reader = ConfigReader::default();
+    let database_factory = DataAccessFactory::new(reader.read_database_info().unwrap(), false)
+        .await
+        .unwrap();
+    let image_review_data = database_factory.get_image_review_data_access();
+    let image_hoster = FlickrApiHandler::new(reader.read_flickr_info().unwrap());
+    let image_scheduling = ImageReviewer::new(image_review_data, image_hoster);
+    Scheduler::new(info, image_scheduling, MealplanManagerMock::default()).await
```
worldofjoni, line 39: My idea was thath we can probably put the tests for all scheduling in a single test file as it is only a single component (we then can still create multiple test cases). Then we would not need a mock.

worldofjoni (changes requested):

worldofjoni:
see overleaf comments

backend/tests/image_review_scheduler.rs
uglrl: I think it its cleaner this way

PR #95, opened 2023-08-22
B/max weeks

worldofjoni:
Parsing one more week from the mensa website

### DIFF
--- a/backend/src/layer/data/database/request.rs
+++ b/backend/src/layer/data/database/request.rs
@@ -8,11 +8,11 @@ use crate::{
         DataError, RequestDataAccess, Result,
     },
     null_error,
-    util::{Additive, Allergen, Date, MealType, Price, Uuid},
+    util::{Additive, Allergen, Date, MealType, Price, Uuid}, layer::data::swka_parser::swka_link_creator::NUMBER_OF_WEEKS_TO_POLL,
 };
 
 /// We have for four weeks, including the current week
-const MAX_WEEKS_DATA: i64 = 4;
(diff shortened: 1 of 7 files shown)


backend/src/layer/data/database/request.rs
```diff
@@ -428,13 +429,13 @@ mod tests {
         assert_eq!(meals, provide_dummy_meals());
 
         let meals_in_future = request
-            .get_meals(line_id, Local::now().date_naive() + Duration::weeks(4))
+            .get_meals(line_id, Local::now().date_naive() + Duration::weeks(5))
             .await
```
uglrl, line 433: Replace the 5 with MAX_WEEKS_DATA +1 and 4 with MAX_WEEKS_DATA

worldofjoni: ok

backend/src/layer/data/database/request.rs
```diff
@@ -11,13 +11,12 @@ use crate::{
...
 
-/// We have for four weeks, including the current week
-const MAX_WEEKS_DATA: i64 = 4;
-
 /// Class implementing all database requests arising from graphql manipulations.
 #[derive(Debug)]
 pub struct PersistentRequestData {
     pub(super) pool: Pool<Postgres>,
+    /// Number of weeks, including the current week, we get/have data for.
+    pub(super) max_weeks_data: u32,
```
uglrl, line 19: Wieso ist das pub(super)?

worldofjoni: because only the Factory is allowed to access these attributes, but the Factoriey lies in another file

uglrl (approved):

PR #104, opened 2023-08-29
Made ci faster

uglrl:


.github/workflows/rust.yml
```diff
@@ -92,27 +82,53 @@ jobs:
...
+    - uses: actions/checkout@v3
+    - run: rustup toolchain install stable --profile minimal
+    - uses: Swatinem/rust-cache@v2
+      with:
+        key: "test"
+    - name: Setup database
+      run: |
+        cargo install sqlx-cli --no-default-features --features native-tls,postgres
+        cargo sqlx database setup
+    - name: Check docs
```
worldofjoni, line 131: this more a compilation thing than a test (and thus would e more appropriate in the other task)

.github/workflows/rust.yml
```diff
@@ -92,27 +82,53 @@ jobs:
...
-          cargo +nightly tarpaulin --verbose --all-features --workspace --timeout 120 --out Xml
+          cargo tarpaulin --all-targets --skip-clean --workspace --timeout 120 --out Xml
       - name:                   Upload to codecov.io
         uses:                   codecov/codecov-action@v2
         with:
           token:                ${{secrets.CODECOV_TOKEN}}
           fail_ci_if_error:     true
 
+
+  test:
```
worldofjoni, line 104: we now either need to specify that tests must also pass before merging is allowed (which must be done by an admin somewhere in the repo) or we maybe can create a dummy task which just depends on all tasks that need to be successfull before merging is allowed (needs see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds)

.github/workflows/rust.yml
```diff
@@ -92,27 +82,53 @@ jobs:
...
+      - uses: Swatinem/rust-cache@v2
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          key: "coverage"
```
worldofjoni, line 89: why do we use different cahches?

worldofjoni (changes requested):

.github/workflows/rust.yml
uglrl: Oh, ok. I thought that the command sets up the database and is needed for the following commands to work

.github/workflows/rust.yml
uglrl: You're the admin....

.github/workflows/rust.yml
worldofjoni: I am takling about the "check docs"

.github/workflows/rust.yml
worldofjoni: And i would like to have only one job which does nothing and depends on many children doing the work in parallel (if that works like i imagine)

worldofjoni (approved):

PR #108, opened 2023-09-29
B/file handler

worldofjoni:

Requires #106

## Backend


backend/src/util.rs
```diff
@@ -5,13 +5,20 @@
...
 
-// Uuid type used in multiple places.
+/// Uuid type used in multiple places.
 pub type Uuid = uuid::Uuid;
 
+/// An in memory representation of an image containing pixels.
+pub type ImageResource = DynamicImage;
+
+/// File extension used for images
+pub const IMAGE_EXTENSION: &str = "jpg";
```
Whatsuup, line 20: If we want to change that extension in production, is it possible that "older" images (*.jpg) can be processed with the new type? And if not, is there a generalized image type, that would prevent such issue?

backend/src/layer/data/file_handler/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+impl ImageStorage for FileHandler {
+    async fn save_image(&self, id: Uuid, image: ImageResource) -> Result<()> {
+        let mut file_path = self.image_path.clone();
+        file_path.push(id.to_string());
+        file_path.set_extension(IMAGE_EXTENSION);
+
+        tokio::task::spawn_blocking(move || image.save(file_path))
+            .await
+            .expect("image saving should not panic nor get aborted")?;
+
```
Whatsuup, line 45: If an image could not be saved (with `image.save(file_path)`), does the server crash? And if not, all operations that follow will be canceled? Is the image_id then referring to nothing? Does the user/client get an error message?

backend/src/util.rs
worldofjoni: the webserver will serve images as they are in the directory but you would need to specify the correct extension on request. when saving the image it will be saved with the specified format/extension. Our image Type is independent of encoding (how it is stored on disk).

backend/tests/command.rs
```diff
@@ -113,17 +113,21 @@ async fn test_remove_image_downvote() {
         .unwrap();
 }
 
+// todo redo / remove test?
```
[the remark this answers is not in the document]
worldofjoni: what do you mean by that?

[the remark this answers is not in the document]
Whatsuup: Removing tests → line coverage degrease.
Why should we remove this test?

backend/src/layer/data/file_handler/mod.rs
Whatsuup: Which reverse seems to be easier? Or should this case be ignored?

backend/src/util.rs
Whatsuup: Ok. Then I assume we never change the extension, as we then can't request all the existing images.

backend/src/layer/data/file_handler/mod.rs
worldofjoni: Probably easier to remove database entry, but in any case, this has to be taken care of in the command layer

backend/src/util.rs
worldofjoni: We could add a database attribute for that but we can also do then when we really want to change filetypes in the future.

backend/tests/command.rs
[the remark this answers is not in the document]
worldofjoni: This was an manual integration test which does not work any longer and does not count to line coverage

Whatsuup (approved):

Issue #124, opened 2023-11-09
Nutrition stats


muety:
The app looks amazing, thanks a lot! As an additional feature, it would be super cool to also see nutrition stats for each meal, see [here](https://github.com/muety/kit-mensa-scraper/tree/master/archive).

worldofjoni:
Thanks! We considered adding nutritinal information but decided not to because in our experience they semed not that realistic/accurate and thus not worth our time. If anyone would like to implement that though it would be a nice addition and should be not to hard to add.

uglrl:
The trawling aspect is now implemented. The database, graphql and app still need to be changed.

worldofjoni:
@uglrl nice, do you want to create a pull request?

uglrl:
No, because the database etc. still has to be implemented. And the trawling still has some bugs

muety:
Would love to have this included! @uglrl if you manage to finish up the data scraping (you might take inspiration from [here](https://github.com/muety/kit-mensa-scraper/blob/a596d037998cb49d513c7e97f376eac8fe8f36c7/lib/src/main/java/edu/kit/aifb/atks/mensascraper/lib/KITMensaScraper.java#L149)) (and possibly also adapt the database schema), I'll probably give it a go and try my hand at the GraphQL API and app integration.

worldofjoni:
Great that you want to contribute! 😄 

I think the following things would be neccesarry to get this going:
Backend:
- [ ] Web scraping: looks like its already implemented by @uglrl , just needs to be reviewed/merged
- [ ] Database schema update incl. access component in backend: I think I @worldofjoni can take a look at that
- [ ]  Changes to the Graphql api. That should not be too much work and a task to get familiar with the project @muety

Frontend:
- [ ] GraphQL-CLient part
- [ ] Internal shananigans like storage to the database for offline use etc. I am not an expert at this part. Maybe @AlexKutschera can help here?
- [ ] Addition to the UI

---
Here are some excerpts of our internal documentation (which we should incoperate into this repository at some point).
## Backend
a bit outdated (especially the image-management part) but the components relevant to us now should still be correct: GraphQL and Database
![image](https://github.com/kronos-et-al/MensaApp/assets/47496388/1f695fd9-2db5-4ad6-851e-64d0ce351ea4)

## Frontend
this might also be dated but nevertheless
[...]

PR #128, opened 2024-01-23
Nutrition data and Environment Data parsing

uglrl:
# Pull Request
Added Nutrition Data
## Backend


worldofjoni (approved):
Thanks for creating the PR.
Looks good to me.

We have to see how we handle these new lints causing errors. Either we disable them or we fix what they tell us (if they are not too much work and don't say anything stupid I think I would prefer the latter)

That is some progress for #124

worldofjoni:
One further question: @uglrl did you include that in the test cases? It looks like but I want to make sure.

Whatsuup:
Looks good to me as well.

I would suggest parsing more than just the nutrition stats, as we're here:

- CO2 emission score
- Water consumption score
- Animal welfare score
- Rainforest suffer score

Tell me how you think about that idea.

[https://www.sw-ka.de/de/hochschulgastronomie/qualitaet_nachhaltigkeit/nachhaltigkeit/](url)

backend/migrations/20230703130438_food.up.sql
```diff
@@ -1,8 +1,8 @@
 -- Add up migration script here
-CREATE TYPE meal_type AS ENUM ('VEGAN','VEGETARIAN','BEEF','BEEF_AW','PORK','PORK_AW','FISH','UNKNOWN');
```
worldofjoni, line 2: You are not allowed to change old database migrations. Otherwise, they cannot be applied to the production database.
If you want to change something, you would need to create a new migration and use SQL functionality to rename tables etc. However, this is a hassle most of the time. Why do you want to rename this?

backend/src/layer/data/swka_parser/test_data/test_1.html
```diff
@@ -1161,8 +1161,8 @@ <h1 class="mensa_fullname">Menseria Moltkestraẞe 30</h1>
 
 
                 <div class="col-4 col-medium-4 col-small-12">
-                  <div><select id="meal_type" name="meal_type" class="p-dd"
```
worldofjoni, line 1164: did you run the tests after renaming? I do not think this should be renamed. Renaming is dangerous.

backend/src/layer/trigger/api/types/meal.rs
```diff
@@ -105,7 +105,7 @@ struct Ratings {
     /// The average rating of this meal.
     average_rating: f32,
     /// The total number of ratings for this meal.
-    ratings_count: u32,
```
worldofjoni, line 108: You are sadly also not allowed to change this, as this is part of the public API. Changing this would break the clients.

worldofjoni (changes requested):
I think it is not worth it to change all those names that have an external dependency now (database, API) as it is just too much of a hassle and too dangerous (that can break things).

backend/migrations/20230703130438_food.up.sql
uglrl: To make it consistent with the frontend and to generalise the concept of a meal type, because sides also have a type

backend/src/layer/data/swka_parser/test_data/test_1.html
uglrl: Yes, I did, they all passed. I thought that we were going to update the database anyway, so I changed it. Clippy was complaining...

worldofjoni: but what you changed here is the HTML of the mensa webpage...

backend/migrations/20230703130438_food.up.sql
worldofjoni: You are right, that would be the better name

backend/src/layer/trigger/api/types/meal.rs
uglrl: Ok. I reverted it

backend/src/layer/data/swka_parser/test_data/test_1.html
uglrl: I know. I fixed it now

backend/migrations/20230703130438_food.up.sql
uglrl: I changed it back in the api. Everywhere else is now adjusted. For the future, we could update the app to use both and when 95% (or more) of people have the new app, then we can change the API. That way everything is consistent between front and backend and there are very few people, who notice a change

backend/src/layer/trigger/api/types/meal.rs
uglrl: @Joni, could you fix the clippy warning in trigger/api/query? And maybe make it so the CI-Pipeline ignores the warnings in trigger/api/types/meal?

worldofjoni: How is the warning called exactly? you can put `#[allow(clippy:xxx)]` around the struct to allow ignore the warning (allow the "problem") there.

backend/migrations/20230703130438_food.down.sql
```diff
@@ -1,4 +1,4 @@
 -- Add down migration script here
 DROP TABLE food;
 
-DROP TYPE meal_type;
```
worldofjoni: don't change old migration, create new one where you rename

backend/migrations/20240214154028_food_env_score.up.sql
```diff
@@ -0,0 +1,17 @@
+-- Add up migration script here
+
+CREATE DOMAIN env_rating as smallint CHECK (VALUE > 0);
+CREATE DOMAIN env_value as smallint CHECK (VALUE > 0);
+
+CREATE TABLE food_env_score (
```
worldofjoni: what were your considerations for creating a new table here?
I think this is right/fine but I would like to hear some thoughts

backend/migrations/20240214154028_food_env_score.up.sql
```diff
@@ -0,0 +1,17 @@
+-- Add up migration script here
+
+CREATE DOMAIN env_rating as smallint CHECK (VALUE > 0);
```
worldofjoni: why do you use `smallint`s here?
In Rust you use `u32` which is a sensible default for an unsigned value, but smallints are only 2 bytes[1]. I think integer does fine here.

[1] https://www.postgresql.org/docs/current/datatype-numeric.html

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -112,23 +115,28 @@ lazy_static! {
     static ref ENV_SCORE_NODE_CLASS_SELECTOR: Selector = Selector::parse("div.enviroment_score.average").expect(SELECTOR_PARSE_E_MSG);
 
     /// A Regex for getting prices in euros. A price consists of 1 or more digits, followed by a comma and then exactly two digits.
-    static ref PRICE_REGEX: Regex = Regex::new(r"([0-9]*),([0-9]{2})").expect(REGEX_PARSE_E_MSG);
+    static ref DECIMAL_REGEX: Regex = Regex::new(r"([0-9]*),([0-9]{2})").expect(REGEX_PARSE_E_MSG);
```
worldofjoni, line 118: I think this rename can lead to dangerous futures: is it clear that a "decimal" always has exactly two decimal digits? This is an essential requirement for the price parsing below in rust code.
I think this rename loses that semantic, maybe a decimal does not even need to have exactly two decimal places

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -369,20 +380,132 @@ impl HTMLParser {
...
-        env_score_node
-            .value()
-            .attr(ENV_SCORE_ATTRIBUTE_NAME)?
-            .parse::<u32>()
+    fn get_dish_env_score(dish_node: &ElementRef) -> Option<EnvironmentInfo> {
+        let env_info = EnvironmentInfo {
+            average_rating: Self::get_average_env_score(dish_node)?,
+            co2_rating: Self::get_co2_rating(dish_node)?,
+            co2_value: Self::get_co2_value(dish_node)?,
+            water_rating: Self::get_rating(dish_node, "div.wasser_bewertung")?,
```
worldofjoni, line 396: should these selectors (`div.wasser_bewertung' etc) not rather be constants?

worldofjoni: maybe that would make thinks more unreadable though, just an idea...

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -369,20 +380,132 @@ impl HTMLParser {
...
+            water_value: Self::get_water_value(dish_node)?,
+            animal_welfare_rating: Self::get_rating(dish_node, "div.tierwohl")?,
+            rainforest_rating: Self::get_rating(dish_node, "div.regenwald")?,
+            max_rating: Self::get_max_rating(dish_node, "div.regenwald")?,
+        };
+        Some(env_info)
+    }
+
+    fn get_average_env_score(dish_node: &ElementRef) -> Option<u32> {
+        Some(
```
worldofjoni, line 406: if this average is always calculated like that (manually) then we should probably not store it in the database in the first place and rather calculate it on access

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -369,20 +380,132 @@ impl HTMLParser {
...
+                + Self::get_rating(dish_node, "div.wasser_bewertung")?
+                + Self::get_rating(dish_node, "div.tierwohl")?
+                + Self::get_rating(dish_node, "div.regenwald")?)
+                / 4,
+        )
+    }
+
+    fn get_co2_rating(dish_node: &ElementRef) -> Option<u32> {
+        Self::get_rating(dish_node, "div.co2_bewertung")
+            .or_else(|| Self::get_rating(dish_node, "div.co2_bewertung_wolke"))
```
worldofjoni, line 417: what are ☁️ doing here? 😄

backend/src/util.rs
```diff
@@ -180,6 +180,27 @@ pub struct Price {
     pub price_pupil: u32,
 }
 
+/// This struct contains all environmental information. co2 in gramms, water in litres
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EnvironmentInfo {
```
worldofjoni, line 185: is this the right place? or should it go to the parser model?

backend/src/layer/data/swka_parser/html_parser.rs
uglrl: It's a stupid website thing...

backend/src/util.rs
uglrl: I mean, we need it for the graphql thing right?

backend/src/layer/data/swka_parser/html_parser.rs
uglrl: Yes. This is more quick and dirty. I can change it tomorrow

backend/migrations/20240214154028_food_env_score.up.sql
uglrl: Creating a new table means not changuing the old ones

backend/migrations/20230703130438_food.down.sql
uglrl: Is this not the original name? I thought i had changed it back

uglrl:
Ok, implementation and Database is now done! We need some tests now.

backend/src/layer/data/swka_parser/html_parser.rs
worldofjoni: I actually would not just create global constants for constants' sake, especially if those are only used once. That way, you always have to scroll from the usage to the top of the document to find out what the constant is referring to. If it is self-explanatory, I would just put it as a "magic string" where it is used, otherwise maybe declare a constant/variable next to where it is used.

backend/src/util.rs
worldofjoni: For graphql to make this struct quariable we need to create a separate version anyways. Only for enums it is more convenient that way (a hassle to duplicate them)

backend/migrations/20240214192738_food_env_score.up.sql
```diff
@@ -1,17 +1,16 @@
 -- Add up migration script here
 
-CREATE DOMAIN env_rating as smallint CHECK (VALUE > 0);
```
worldofjoni: does this work with `INT`? In postgres it is at least convention to use lowercase for types (`integer`)

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -333,6 +355,36 @@ impl PersistentMealplanManagementData {
         .execute(&self.pool)
         .await?;
 
+        if nutrition_data.is_some() {
+            let nut_data = nutrition_data.expect("If you see this, logic broke");
```
worldofjoni, line 359: The nice way to do this:
```suggestion
        if let Some(nut_data) = nutrition_data {
```

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -333,6 +355,36 @@ impl PersistentMealplanManagementData {
...
+                i32::try_from(nut_data.sugar)? as _,
+                i32::try_from(nut_data.fat)? as _,
+                i32::try_from(nut_data.saturated_fat)? as _,
+                i32::try_from(nut_data.salt)? as _,
+                food_id,
+            ).execute(&self.pool).await?;
+        }
+
+        if environment_information.is_some() {
+            let env_info = environment_information.expect("If you see this, logic broke");
```
worldofjoni, line 374: same as above

backend/src/layer/data/database/request.rs
```diff
@@ -286,6 +286,47 @@ impl RequestDataAccess for PersistentRequestData {
...
+    async fn get_nutrition_data(&self, food_id: Uuid) -> Result<Option<NutritionData>> {
+        let res = sqlx::query!(
+            r#"SELECT energy, protein, carbohydrates, sugar, fat, saturated_fat, salt FROM food_nutrition_data WHERE food_id = $1"#,
+            food_id
+        ).fetch_optional(&self.pool)
+        .await?;
+        if res.is_none() {
+            return Ok(None);
+        }
+        let res = res.expect("isdpoiopvü");
```
worldofjoni, line 299: Maybe something like
```rust
let res = match res {
  Some(res) => res,
  None => return Ok(None)
}
```
Or you could use Option::map but the closure would need to return a Result, so you get Option<Result<...>>. On that, you can call Option::transpose [1] to get Result<Option<..>,..>. I think that is what we want, maybe there is also an early return version using transpose and ?, but things get complicated here, so the match may be the most readable.

[1] https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose

backend/src/layer/data/database/request.rs
```diff
@@ -286,6 +286,47 @@ impl RequestDataAccess for PersistentRequestData {
...
+
+    async fn get_environment_information(&self, food_id: Uuid) -> Result<Option<EnvironmentInfo>> {
+        let res = sqlx::query!(
+            r#"SELECT co2_rating, co2_value, water_rating, water_value, animal_welfare_rating, rainforest_rating, max_rating FROM food_env_score WHERE food_id = $1"#,
+            food_id
+        ).fetch_optional(&self.pool).await?;
+        if res.is_none() {
+            return Ok(None);
+        }
+        let res = res.expect("isdpoiopvü");
```
worldofjoni, line 319: same as above

backend/src/util.rs
```diff
@@ -183,15 +183,13 @@ pub struct Price {
 /// This struct contains all environmental information. co2 in gramms, water in litres
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EnvironmentInfo {
-    /// The number of stars the food has on average. Out of `max_rating`
-    pub average_rating: u32,
     /// The number of stars the food has for co2 emmissions. Out of `max_rating`
     pub co2_rating: u32,
-    /// The amount of co2 emitted by the production of the food
+    /// The amount of co2 emitted by the production of the food. In gramms
```
worldofjoni, line 188:
```suggestion
    /// The amount of CO2 emitted by the production of the food. In grams.
```
same for milliliters. Important information 👍 .

backend/migrations/20240214154028_food_env_score.up.sql
worldofjoni: Furthermore, this allows us to model the food:score relation as 1:[0,1] correctly

backend/src/layer/data/swka_parser/html_parser.rs
worldofjoni: I think this could be best done on database access

backend/src/layer/data/database/mealplan_management.rs
uglrl: Right. Thank you. I forgot that this worked in Rust. It's been a while...

backend/migrations/20240214192738_food_env_score.up.sql
worldofjoni: In other tables we wrote that out to `integer` nor just `int` so I would prefer that for consistancy

worldofjoni (commented):
I see, you already implemented the API changes, nice!
If I see that correctly, everything should be working now except calling the insertion functions into the database (@Whatsuup part), but some tests are still missing.
To to be precise, tests are needed for:
- [ ] Parsing
- [ ] Insertion into database
- [ ] Extraction from database
- [ ] GraphQL API access

am I missing anything? Are any of those done yet?

@uglrl maybe you can also post an excerpt of the changes on the GraphQL schema (accessible in the GraphQL playground web UI) so we can check if things also look fine there?

When all those things are done I will do a final review and I think we can merge then.

Edit:
- [ ] As we changed the API, we should also update the version inside cargo.toml, to `1.1.0` I think.

worldofjoni:
Changes to the Graphql schema:
@AlexKutschera you may want to take a look at that and approve
```graphql

type Meal {
  # The identifier of the main course.
  id: UUID!

  # The name of the main course.
  name: String!
  
  # ...

  # Provides the environment information of this meal.
  environmentInfo: EnvironmentInfo

  # Provides the nutrition data of this meal.
  nutritionData: NutritionData
}

# This struct contains all environmental information. co2 in grams, water in litres
type EnvironmentInfo {
  # The average environmental rating. Out of `max_rating`
  averageRating: Int!

  # The number of stars the food has for CO2 emmissions. Out of `max_rating`
  co2Rating: Int!

  # The amount of CO2 emitted by the production of the food. In grams
  co2Value: Int!

  # The number of stars the food has for water consumption. Out of `max_rating`
  waterRating: Int!

  # The amount of water used for the production of the food. In Millilitres
  waterValue: Int!

  # The number of stars the food has for animal welfare. Out of `max_rating`
  animalWelfareRating: Int!

  # The number of stars the food has for rainforest preservation. Out of `max_rating`
  rainforestRating: Int!

  # The maximum amount of stars for each category
  maxRating: Int!
}

# The nutrients of a dish
type NutritionData {
  # Energy in Kcal
  energy: Int!

  # Protein in grams
  protein: Int!

  # Carbs in grams
  carbohydrates: Int!

  # Sugar in grams
  sugar: Int!

  # Fat in grams
  fat: Int!

  # Saturated fat in grams
  saturatedFat: Int!

  # Salt in grams
  salt: Int!
}

```

[schema.graphq.txt](https://github.com/kronos-et-al/MensaApp/files/14322028/schema.graphq.txt)

AlexKutschera (approved):

backend/migrations/20240214192718_food_nutrition_data.up.sql
```diff
@@ -0,0 +1,16 @@
+-- Add up migration script here
+
+CREATE DOMAIN weight_domain as integer CHECK (VALUE > 0);
+CREATE DOMAIN energy_domain as integer CHECK (VALUE > 0);
```
worldofjoni, line 4: Should they not also be allowed to be zero?

backend/migrations/20240214192738_food_env_score.up.sql
```diff
@@ -0,0 +1,16 @@
+-- Add up migration script here
+
+CREATE DOMAIN env_rating as integer CHECK (VALUE > 0);
+CREATE DOMAIN env_value as integer CHECK (VALUE > 0);
```
worldofjoni, line 4: here also?

backend/migrations/20240214192718_food_nutrition_data.up.sql
worldofjoni: do we redundantly call them  `_domain`? we did not do that for the other domains

backend/src/layer/data/database/request.rs
```diff
@@ -286,6 +286,54 @@ impl RequestDataAccess for PersistentRequestData {
...
+            r#"SELECT co2_rating, co2_value, water_rating, water_value, animal_welfare_rating, rainforest_rating, max_rating FROM food_env_score WHERE food_id = $1"#,
+            food_id
+        ).fetch_optional(&self.pool).await?;
+        if let Some(res) = res {
+            let co2_rating = u32::try_from(res.co2_rating)?;
+            let water_rating = u32::try_from(res.water_rating)?;
+            let animal_welfare_rating = u32::try_from(res.animal_welfare_rating)?;
+            let rainforest_rating = u32::try_from(res.rainforest_rating)?;
+            let average_rating =
+                (co2_rating + water_rating + animal_welfare_rating + rainforest_rating) / 4;
```
worldofjoni, line 322: does that correctly get converted to f32?

backend/src/layer/trigger/api/query.rs
```diff
@@ -64,6 +64,7 @@ impl QueryRoot {
 
     /// This query returns the version of this API schema. It can also be used for health checks.
     #[instrument(skip(self, _ctx))]
+    #[allow(clippy::no_effect_underscore_binding)]
```
worldofjoni, line 67: about what is clippy complaining here? whey do we ignore that?
https://rust-lang.github.io/rust-clippy/master/index.html#/underscore_b

backend/src/util.rs
```diff
@@ -180,6 +180,25 @@ pub struct Price {
...
+    /// Sugar in grams
+    pub sugar: u32,
+    /// Fat in grams
+    pub fat: u32,
+    /// Saturated fat in grams
+    pub saturated_fat: u32,
+    /// Salt in grams
+    pub salt: u32,
+}
+
```
worldofjoni, line 201: why is that here? is that not sufficient in the model?

worldofjoni (approved):

worldofjoni:
Besides these notes, I think we are ready to merge.

worldofjoni:
FYI: I have also ignored the startup files regarding test coverage, so we are now well above 90% again

backend/src/layer/data/database/request.rs
uglrl: No, this should be an unsigned integer, like the original.

backend/src/util.rs
uglrl: Because this is used throughout the backend

backend/src/layer/data/database/request.rs
worldofjoni: Never mind, Github showed me the wrong type for average_rating (the one from Meal). I should trust the type system 😁

backend/src/layer/trigger/api/query.rs
worldofjoni: alright, I think this is a problem with the linter not detecting that this argument is required because of the macro. Anoying but ignoring is right.

backend/src/util.rs
worldofjoni: I remember we had talked about that. My oddity-detection is just triggered that environment info is not her but this is.

uglrl: Environment_info was here, but i moved it to two different places, because there are two variants: One with Average Rating (Pre-Database) and one without (post-database)

worldofjoni: yea, i see

worldofjoni: all these duplicate similar structs are a bit a problem of our code base but i still think they should be seperate structs if they specify seperate things. Maybe this could be solved using a variant of the typestate pattern in another life.
(This would allow **one** struct definition to produce multiple distinct types---one with an attribute and one without---that can be differentiated at compile time)

PR #132, opened 2024-02-17
Update rust.yml

uglrl:
Renewed workflow for backend


worldofjoni:
I want to run the workflow successfully before merging that. Therefore we have to wait for #128 as the new lints require changes done there so that the workflow is successful.

worldofjoni (changes requested):
Reminder: I also want to do package updates (and security updates) here

PR #133, opened 2024-02-20
Merge nutrition resolving into nutrition feature branch

Whatsuup:
## Backend


Whatsuup:
Already done by @uglrl. 
I opened this PR without checking the current status.
I also couldn't estimate how much work it would be. It was only 4 lines of code. 
Nevertheless, thank you @uglrl!

To do at least something, I ran the migration tests and checked them randomly. I didn't notice anything negative. Well done.

worldofjoni:
@Whatsuup are the database insertion methods also called (at the correct place)? I am not aware that that is already done.

Whatsuup:
Added env and nutrition update feature

Whatsuup:
> @Whatsuup are the database insertion methods also called (at the correct place)? I am not aware that that is already done.

backend/src/layer/data/database/mealplan_management.rs
Whatsuup, line 1: SQL-Updates can be joined but I don't think this is a performance issue.

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -862,23 +955,68 @@ mod test {
...
+        .await
+        .unwrap();
+        let db_res_rating = selections.first().unwrap().rainforest_rating;
 
-        assert_eq!(selection.name, name);
+        assert_eq!(db_res_name, name);
+        assert_eq!(db_res_rating, environment_info.rainforest_rating as i32);
+        assert_eq!(db_res_salt, nutrition_data.salt as i32);
+        // tbd: checking env and nutrition data
+        // This could be done better as it is now.
```
Whatsuup, line 1019: There is a way to get complete structs without manually parsing out of the query (query-scalar?). This is something that I want to change for this and all other related tests.

backend/src/layer/data/database/mealplan_management.rs
worldofjoni: you mean because there are separate SQL calls? this is not really avoidable and ok.

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -267,18 +281,96 @@ impl MealplanManagementDataAccess for PersistentMealplanManagementData {
...
+        self.update_environment_data(&food_id, environment_info)
+            .await?;
 
         Ok(())
     }
 
+    async fn update_nutrition_data(
+        &self,
+        food_id: &Uuid,
+        nutrition_data: Option<NutritionData>,
```
worldofjoni, line 312: maybe move the option out of here and call this function conditionally?
This would resemble the semantics more that nothing happens if no new data is provided.
(Is this what we want? if one time the nutrition data is forgotten, we keep the old one; but we can never revert to "no data"---I think this is what we want)

backend/src/layer/data/database/mealplan_management.rs
worldofjoni: Do you mean `query_as`?
https://docs.rs/sqlx/latest/sqlx/macro.query_as.html

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -946,6 +1084,7 @@ mod test {
                 .await
                 .unwrap();
         assert_eq!(&actual_name, name);
+        // tbd: checking env and nutrition data
```
worldofjoni, line 1087: what needs to be done here?

worldofjoni (approved):
Good!
you need to run `sqlx prepare` so that the ci can run successfully

backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -1313,4 +1201,94 @@ mod test {
         assert_eq!(price.price_guest, record.price_guest as u32);
         assert_eq!(price.price_pupil, record.price_pupil as u32);
     }
+
+    async fn get_env_info(pool: &PgPool, food_id: Uuid) -> ParseEnvironmentInfo {
+        ParseEnvironmentInfo::from_row(
+            &sqlx::query(&format!(
+                r#"SELECT * FROM food_env_score WHERE food_id = '{food_id}'"#
```
worldofjoni, line 1208: just don't... not even in tests...
Why not useing bind?

Whatsuup: I don't get the Problem. What should I "bind" there?

worldofjoni: You are formatting a value (input) into the sql string. If this would not be a test and maybe a string input, this would create a SQL-injection vulnerability.
To do it properly, you would add a placeholder `#1`

PR #139, opened 2024-02-27
Poultry recogniser

uglrl:


AlexKutschera (commented):
Ich glaube es braucht noch eine Migration um den meal_type in der Datenbank anzupassen. Siehe https://github.com/kronos-et-al/MensaApp/blob/5ce67d7a64e18f1828f90c0cf8d74f320fae9449/backend/migrations/20230703130438_food.up.sql#L2. Ansonsten sieht das denke ich gut aus

uglrl:
Ich hab das jetzt mal in die Datenbank hinzugefügt. Jetzt muss das "nur" noch im frontend angepasst werden.

backend/migrations/20230703130438_food.up.sql
```diff
@@ -1,5 +1,5 @@
 -- Add up migration script here
-CREATE TYPE meal_type AS ENUM ('VEGAN','VEGETARIAN','BEEF','BEEF_AW','PORK','PORK_AW','FISH','UNKNOWN');
```
worldofjoni, line 2: As already stated last time, you are not allowed to edit existing, already deployed migrations. Rather create a new migration where you edit the enum.

worldofjoni: In this new migration, you should also modify all existing meals where the poultry regex matches the new meal type, so new parsed meals can get merged with the existing ones.

worldofjoni:
Some statistics on the poultry regex matches on the existing unknown meals:
We currently have 285 main dishes with unknown type (out of 1363 mains total).
178 of them match with the new regex.
Of the remaining 107, most are either wrongly missing a category or have some sort of fish that maybe does not count under the mensa's fish category.

We also have "Hausgeräucherte Entenbrust mit Mango - Chili - Chutney" which should also include according to Wikipedia[1].
Some more meals we could include: 
- BBQ Chicken Wings Vollkorn Naturreis in Paprikasoße
- Chicken Drum Sticks
So we should also add `chicken` as a keyword
[1] https://en.wikipedia.org/wiki/Poultry

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -133,6 +133,8 @@ lazy_static! {
     static ref VOLUME_REGEX: Regex = Regex::new(r"([0-9]*),([0-9]{2}) l").expect(REGEX_PARSE_E_MSG);
 
     static ref ID_REGEX: Regex = Regex::new(r"[0-9]{18,}").expect(REGEX_PARSE_E_MSG);
+
+    static ref POULTRY_REGEX: Regex = Regex::new(r"(Pute|(G|g)eflügel|H(ü|ä|u|a)hn)").expect(REGEX_PARSE_E_MSG);
```
worldofjoni, line 137: I think we should add the regex i flag to ignore case. This way we have no problems if these keywords appear in the middle of a sentence.
https://docs.rs/regex/latest/regex/struct.RegexBuilder.html#method.case_insensitive

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -316,12 +318,13 @@ impl HTMLParser {
...
     fn get_dish(dish_node: &ElementRef) -> Option<Dish> {
+        let name = Self::get_dish_name(dish_node)?;
         Some(Dish {
-            name: Self::get_dish_name(dish_node)?,
+            name: name.clone(),
             price: Self::get_dish_price(dish_node),
             allergens: Self::get_dish_allergens(dish_node).unwrap_or_default(),
             additives: Self::get_dish_additives(dish_node).unwrap_or_default(),
-            food_type: Self::get_dish_type(dish_node).unwrap_or(FoodType::Unknown),
+            food_type: Self::get_dish_type(dish_node, &name).unwrap_or(FoodType::Unknown),
```
worldofjoni, line 327: maybe you can reorder these attributes so you do not need to clone `name`

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -397,12 +400,28 @@ impl HTMLParser {
...
+                .attr(DISH_TYPE_ATTRIBUTE_NAME)
+                .map(FoodType::parse);
+            if preliminary_dish_type.is_none() || Some(FoodType::Unknown) == preliminary_dish_type {
+                return Self::check_for_poultry(name);
+            }
+            preliminary_dish_type
+        } else {
+            Self::check_for_poultry(name)
+        }
+    }
```
worldofjoni, line 417: I may be getting too critical, but this seems more nested than necessary.
Maybe with `.and_then`s this can all be made linear

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -397,12 +400,28 @@ impl HTMLParser {
...
+
+    fn check_for_poultry(name: &str) -> Option<FoodType> {
```
worldofjoni, line 419: At least regarding to the function's name, I think returning a `bool` would be more idiomatic.

worldofjoni (changes requested):

AlexKutschera (approved):

backend/migrations/20240313130248_poultry_checker.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add up migration script here
+ALTER TYPE meal_type add value 'POULTRY';
+COMMIT;
+
+UPDATE food SET food_type = 'POULTRY' WHERE name ~ '(?i)ente|chicken|pute|gefl[üÜ]gel|h[üäÜÄua]hn';
```
worldofjoni, line 5: no upper case letters in regex needed

backend/migrations/20240313130248_poultry_checker.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add up migration script here
+ALTER TYPE meal_type add value 'POULTRY';
+COMMIT;
```
worldofjoni, line 3: is this commit necessary/does it do anything?

backend/migrations/20230703130438_food.up.sql
worldofjoni, line 2: I do not think you are even allowed to insert whitespace, as the files hashes are compared.

backend/src/layer/data/swka_parser/html_parser.rs
```diff
@@ -133,6 +133,8 @@ lazy_static! {
     static ref VOLUME_REGEX: Regex = Regex::new(r"([0-9]*),([0-9]{2}) l").expect(REGEX_PARSE_E_MSG);
 
     static ref ID_REGEX: Regex = Regex::new(r"[0-9]{18,}").expect(REGEX_PARSE_E_MSG);
+
+    static ref POULTRY_REGEX: Regex = Regex::new(r"(?i)ente|chicken|pute|gefl[üÜ]gel|h[üäÜÄua]hn").expect(REGEX_PARSE_E_MSG);
```
worldofjoni, line 137: Unnecessary upper case letters

worldofjoni (approved):
Sorry, it took me some time, but here we are

backend/migrations/20240313130248_poultry_checker.up.sql
uglrl: Yes, it ensures that the previous operation is finished, before the next one is started.

worldofjoni:
Note to myself @worldofjoni 
When deploying this, the database needs to be migrated

AlexKutschera:
Es fehlt glaube ich noch der Geflügel Typ im GraphQL Schema:
https://github.com/kronos-et-al/MensaApp/blob/f2ec05ac935773cb333cbfdcddfba97cff6433ec/backend/src/util.rs#L127

worldofjoni:
> Es fehlt glaube ich noch der Geflügel Typ im GraphQL Schema:
No. You are referencing the wrong version.
https://github.com/kronos-et-al/MensaApp/blob/ab985dcaa59909fe2a847bd2f6d2c8e50aaaf917/backend/src/util.rs#L143C1-L143C13

PR #141, opened 2024-03-04
Update actions

uglrl:
Updated outdated actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,8 +19,8 @@ jobs:
         working-directory: ./app
 
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-java@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-java@v4
       with:
         distribution: 'zulu'
         java-version: '11'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,11 +2,11 @@ name: Build-Android
 
 on:
   push:
-    branches: [ "main", "frontend" ]
+    branches: [ "main"]
     paths: 
       - "app/**"
   pull_request:
-    branches: [ "main", "frontend" ]
+    branches: [ "main"]
     paths: 
       - "app/**"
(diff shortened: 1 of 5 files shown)


uglrl:
I fixed the flutter version in the main.yml file. Now the workflow fails, because the tests are failing. It would be nice if someone could take a look at them

.github/workflows/main.yml
```diff
@@ -2,11 +2,11 @@ name: Build-Android
...
   push:
-    branches: [ "main", "frontend" ]
+    branches: [ "main"]
     paths: 
       - "app/**"
   pull_request:
-    branches: [ "main", "frontend" ]
+    branches: [ "main"]
     paths: 
       - "app/**"
```
worldofjoni, line 11: when removing this here, this action should run on all PRs into main, regardless whether something changed in app

.github/workflows/rust.yml
```diff
@@ -2,11 +2,11 @@ name: Build-Rust
...
   push:
-    branches: [ "main", "backend" ]
+    branches: [ "main"]
     paths: 
       - "backend/**"
   pull_request:
-    branches: [ "main", "backend" ]
+    branches: [ "main"]
     paths: 
       - "backend/**"
```
worldofjoni, line 11: same here

worldofjoni (approved):

.github/workflows/main.yml
```diff
@@ -2,13 +2,11 @@ name: Build-Android
...
   pull_request:
-    branches: [ "main", "frontend" ]
-    paths: 
-      - "app/**"
+    branches: [ "main"]
   workflow_dispatch:
   
 jobs:
```
worldofjoni, line 12: we should rename the build jobt to maybe `build_flutter` because it currently is ambiguous with the secondary build job required for build_rust. On the merge requirements, I can only specify a name of a job. Currently, when specifying build, the backend build job is required redundantly (as it is also a dependency of build_rust)

worldofjoni (approved):

PR #148, opened 2024-04-05
correct timezone for schedules

worldofjoni:
Timezone of parse schedules is now queried from OS on startup.


worldofjoni:
Note to myself @worldofjoni:
do not forget to change the parse time in the production install

uglrl (approved):

PR #151, opened 2024-04-19
Added ability to run without ext. services

uglrl:


worldofjoni (commented):
Nice that you worked on that @uglrl!
I think it would be cleaner to add a command line option to the existing executable for running in isolated _demo mode_ as this is mainly a code duplication. Similar to how the help page or `should_migrate` is done.
That could be tricky, however, because of the generic type parameters.

uglrl:
I thought about doing that, but decided against it, since it would make the normal code flow more difficult to read and understand.

worldofjoni:
Alright. I think, however, that naming the fine just `bin.rs` inside the startup folder is very misleading---something like `standalone_demo_bin.rs` would be clearer.
Also, the usage of this standalone version should be mentioned in the backend readme.

worldofjoni (changes requested):
Alright. I think, however, that naming the fine just `bin.rs` inside the startup folder is very misleading---something like `standalone_demo_bin.rs` would be clearer.
Also, the usage of this standalone version should be mentioned in the backend readme.

worldofjoni (approved):

PR #154, opened 2024-04-19
geflügel aus dem gerichtsnamen erkennen

worldofjoni:
Second PR for this after #129 because of human error.
Waiting for frontend part to finish.


AlexKutschera (approved):

app/assets/locales/de/filter.json
```diff
@@ -7,6 +7,7 @@
   "foodTypeSelectionUnknown": "Keine Angabe",
   "foodTypeSelectionPork": "Schweinefleisch",
   "foodTypeSelectionBeef": "Rindfleisch",
+  "foodTypeSelectionPoultry": "Geflügelfleisch",
```
worldofjoni, line 10: I think it technically is only "Geflügel"

worldofjoni (commented):
@AlexKutschera from my part this is ready

PR #157, opened 2024-04-20
Nährwertangaben

AlexKutschera:
Die App zeigt Nährwertangaben an


app/assets/locales/de/allergen.json
```diff
@@ -1,5 +1,5 @@
 {
-  "allergenTitle": "Allergene:",
```
worldofjoni, line 2: maybe also rename the key?

app/assets/locales/de/nutritionData.json
```diff
@@ -0,0 +1,17 @@
...
+  "carbohydrates": "Kohlenhydrate",
+  "carbohydratesUnit": "g",
+  "sugar": "davon Zucker",
+  "sugarUnit": "g",
+  "fat": "Fett",
+  "fatUnit": "g",
+  "saturatedFat": "davon gesättigt",
+  "saturatedFatUnit": "g",
+  "salt": "Salz",
+  "saltUnit": "g"
```
worldofjoni, line 16: Wouldn't it be better practice to only have one entry for grams? This way, if one unit changes in the future, a change has to be made in the source code but (possibly) not here. This is also kind of redundant.

worldofjoni (approved):
As far as I can tell, this looks fine to me.
@muety do you approve how your changes are incorporated into this?

app/assets/locales/de/nutritionData.json
uglrl: I think, that it is useful to keep the different units separated. Anyway, I don't think that this is a big issue nontheless, since it is highly unlikely that the nutrition data table set by the EU will change much in future.

muety:
> As far as I can tell, this looks fine to me. @muety do you approve how your changes are incorporated into this?

Approved :+1:

app/assets/locales/de/allergen.json
AlexKutschera: maybe, but as i might have changes intersecting between this branch and the envinfo branch i won't change it now to not break anything when merging

PR #164, opened 2024-06-19
Better Reports 1/2

worldofjoni:
Completely overhauled reports (#160), part one of two:
- ergonomic HTML mails with much more information
- better mail subject 
- mail thread grouping for reports of the same image

Still missing (version two):
- one-click delete/verify of image

## Example
image

Issue #160, opened 2024-05-08
More ergonomic image report reviews


worldofjoni:
**Is your feature request related to a problem? Please describe.**
Handling image reviews at the moment consists of scrolling through a list of emails, looking at each reported image and if appropriate, sending a SQL command to the database to hide that image manually. Furthermore, multiple reports for the same image are not easily recognizable and create more work.

**Describe the solution you'd like**
Refine report handeling process:
- Grouping of reports for the same image in the mail program (same subject or setting an ID?)
- adding image ID to mail subject
- let mail subject reflect if image got hidden automatically
- webhook with authentication cookie to easily hide an image
- add meal name to mail, so it can if an image contains the wrong meal. Other images would also be nice.
- maybe embed an image into mail rather than only a link. Might be problematic, As one may not want inappropriate images in their mailbox
- provide a nicer HTML mail template

**Describe alternatives you've considered**
An admin UI. This is a future goal as it is much more work. Meanwhile, nicer emails would be an improvement.

**Additional context**
As of now, there have been 221 reports since October 2023.
They have a format like [this](https://github.com/kronos-et-al/MensaApp/blob/43cccc038a4cc48df39b59311fdc10ecd57f245f/backend/src/layer/data/mail/template.txt)

PR #166, opened 2024-07-08
B/better reports 2

worldofjoni:
Second phase of better image report handling (#160), follow up to #164.

Contains:
- RESTful admin API under `/admin` with BASIC auth
- API endpoints to delete and verify images
- one-click buttons in report mail to delete/verify image

TODO
- [x] human-readable admin API response
- [ ] ~~improve notification email for when image is deleted / verified (same content as report mail?)~~
- [x] tests for database access
- [x] test for image file deletion
- [x] test for command
- [x] maybe tests for admin API?


backend/README.md
```diff
@@ -5,6 +5,17 @@ Backend application for providing and synchronizing meal plan data of the cantee
 
 If you just want to use the (Android, iOS) App, the following is not necessary.
 
+
+## APIs
+
+There are two kinds of APIs available:
+- The main GraphQL API for accessing data like meal plans etc. \
+  This API is accesssable under `/` and ad documentation can be accessed on the there hosted GraphQL playground.
```
Whatsuup, line 13:
```suggestion
  This API is accesssable under `/` and the documentation can be accessed on the there hosted GraphQL playground. 
```

backend/src/layer/trigger/api/admin.rs
```diff
@@ -0,0 +1,105 @@
...
+) -> impl IntoResponse {
+    let Some(creds) = creds else {
+        return Err(unauthenticated());
+    };
+
+    if creds.0 .0.username() != ADMIN_USER || creds.0 .0.password() != auth_key.0 {
+        return Err(unauthenticated());
+    }
+
+    Ok(next.run(req).await)
```
Whatsuup, line 104: project structure violation: trigger layer should not interact with external sources.
This is done by the data layer. Trigger should only prepare user interaction for the logic layer. 
If I get this right, more than that is done here.

Whatsuup:
How about a rest api data layer component? (Should be done in another pr)

Whatsuup (commented):
Better REST API responses in browser

Whatsuup (approved):

PR #177, opened 2025-01-22
B/nutri fix

### LINKED ISSUE #176: Nutrition stats missing
Thankfully, nutrition stats were introduced in #124 to the app. However, I noticed that they're missing quite frequently, even though available on the official website. I can't provide a clear way to "reproduce" this, but here's an exemplary screenshot. I'd say this happens almost daily for a subset of meals. Perhaps something's wrong with the scraper?

worldofjoni:
fixes #176. Nutri/env data entries did not get created on update properly.


backend/src/layer/data/database/mealplan_management.rs
```diff
@@ -309,17 +309,18 @@ impl PersistentMealplanManagementData {
     }
 
     async fn update_nutrition_data(&self, food_id: &Uuid, data: NutritionData) -> Result<()> {
+        // Newly inserting nutrition data is also possible!
         sqlx::query!(
             "
-            UPDATE food_nutrition_data 
+            INSERT INTO food_nutrition_data (food_id, energy, protein, carbohydrates, sugar, fat, saturated_fat, salt) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+            ON CONFLICT (food_id) DO UPDATE
```
Whatsuup, line 316: "ON CONFLICT (data) DO UPDATE" is similar to SQL-Syntax "ON DUPLICATE KEY UPDATE (data)"?

Whatsuup (approved):

backend/src/layer/data/database/mealplan_management.rs
worldofjoni: Yes, this is the postgres version of that.

PR #178, opened 2025-01-22
added poultry parsing

worldofjoni:
added parsing of polutry meal type


backend/src/layer/data/swka_parser/html_parser.rs
uglrl, line 1: You could just rename the get_preliminary_food_type to get_food_type and delete the old get_food_type function

uglrl (approved):

PR #182, opened 2025-02-24
Fix wrong app UI colors.

worldofjoni:
Because the theme color `background` is deprecated, `surface` was used instead, which is a different color and made the entire background too dark/bright.

Also updated flutter versions and dependencies.

Also: updating the meal plan (swiping down) is slow because of database inserts, not because of API queries. There are tons of database queries required for inserting, which i think (hope) could be bundled together better.


app/pubspec.yaml
```diff
@@ -19,8 +19,8 @@ publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 version: 1.0.4+15
```
worldofjoni, line 19: do I need to change this? just bump to +16 ?

AlexKutschera (commented):
Is this not a duplicate of #189

worldofjoni:
already merged with #189

PR #191, opened 2025-04-06
Image validation improvements

### LINKED ISSUE #185: Check image appropriateness using the free Gemini API
Images are often unrelated to the actual food.

The images along with the description of the food could be sent to the the free Gemini 2.0 Flash API to check whether they match. If they don't, reject the image. Perhaps the reason the LLM provides could be shown to the user.

From my testing, Gemini is reasonably good at this task. I used the following prompt, which could be further refined: `Bildet das Bild folgendes ab: "<Beschreibung des Gerichts>"? Antworte mit "Ja" oder "Nein", gefolgt von einer kurzen Begründung.`

The rate limit is 15 requests per minute and 1500 requests per day. I assume this should be plenty.

Whatsuup:
# Pull Request
- [x] Api Handler split
- [x] Image evaluation split
- [x] Json-Structs for requests & responses
- [x] Proper error handling + more error details
- [x] Gemini implementation
- [x] Api-Info adjustment
- [x] .env adjustments
- [x] secret updates for new api tests

## Backend


Whatsuup:
Here are some testing results. This test should decide whether the Gemini-API turns out to be useful.
As far as I can say the results are not optimal but considering the temporary state during rush hours, it seems to be much better than now.

image

Just a small change in the sentence or one simple word more in the request improves the outcome by ~15%!

image

Almost all invalid images are detected by AI. Images an admin would accept do not score as well with the AI. More on that in the next diagram.
For now, the AI seems much striker than a human.

image

When we look at the reasons why Gemini rejects an image, some concerns arise. Even if sometimes just one ingredient is missing or the egg is boiled instead of cooked, the AI ​​rejects the image. 
Sometimes Gemini can't differentiate or detect the right food. For example: Kartoffelsalat is incorrectly recognized as Kartoffelbrei.
Some custom meal names like "koerifries" provide issues, as it's hard to know what "koerifries" are. The AI expects curry sauce or something similar with the fries.
Finally, there are issues with the naming of the meals. Meals like "Buffet with x, y, z" cause problems because Gemini expects x, y, and z, even when there's a buffet selection. Homemade mensa noodles aren't recognized as homemade by the AI ​​because they look industrial. The dish should contain cashews, but the cafeteria doesn't provide enough, so they're not visible in the image.
I guess you can see were this is going...

But despite these issues, I see a lot of potential and am implementing the API for our purposes. For modularity and future improvements, I'll add .env variables for the request body and specify whether to use Gemini at all.

backend/src/layer/data/image_validation/google_api_handler.rs
```diff
@@ -18,16 +20,26 @@ use std::io::Cursor;
 ///     - calling the api<br>
 ///     - determine which images are allowed<br>
 ///     - and returning helpful error messages if an image got not accepted
+#[derive(Default)]
```
worldofjoni, line 23: Seeing that, I am considering wheter we should separate these handlers completely, making them separate components in the data layer using the same interface... But maybe that is overkill.

backend/src/layer/data/image_validation/google_api_handler.rs
```diff
@@ -36,19 +48,53 @@ impl GoogleApiHandler {
...
 #[async_trait]
 impl ImageValidation for GoogleApiHandler {
     async fn validate_image(&self, image: &ImageResource) -> Result<()> {
-        let b64_image = image_to_base64(image)?;
-        let results = self.request.encoded_image_validation(b64_image).await?;
-        self.evaluation.verify(&results)
+        let mut safe_search_result = Ok(());
+        let mut gemini_result = Ok(());
+        if let Some(handler) = self.safe_search_handler.as_ref() {
+            let b64_image = image_to_base64(image)?;
```
worldofjoni, line 82: only needs to be done once if both are enabled

backend/src/layer/data/image_validation/google_api_handler.rs
```diff
@@ -36,19 +48,53 @@ impl GoogleApiHandler {
...
 }
 
 #[async_trait]
 impl ImageValidation for GoogleApiHandler {
     async fn validate_image(&self, image: &ImageResource) -> Result<()> {
-        let b64_image = image_to_base64(image)?;
-        let results = self.request.encoded_image_validation(b64_image).await?;
-        self.evaluation.verify(&results)
+        let mut safe_search_result = Ok(());
+        let mut gemini_result = Ok(());
```
worldofjoni, line 80: These look wiered, maybe there is a cleaner way to do what you want to do.

backend/src/layer/data/image_validation/google_api_handler.rs
worldofjoni: thus, it is probably better to do both in the same component (ref above).

backend/src/layer/data/image_validation/gemini_validation/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod gemini_evaluation;
+pub mod gemini_request;
+/// This module contains all component related structs that are used to compile or decompile json.
```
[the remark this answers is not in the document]
worldofjoni: mayb not, idk.

backend/src/startup/config.rs
```diff
@@ -232,25 +232,50 @@ impl ConfigReader {
...
     }
 }
 
 fn read_var(var: &str) -> Result<String> {
     env::var(var).map_err(|e| ServerError::MissingEnvVar(var.to_string(), e))
 }
 
-fn read_acceptence_var(key: &str) -> Result<[u8; 5]> {
+fn read_var_to_bool(key: &str) -> Result<bool> {
+    match read_var(key)?.to_lowercase().as_str() {
```
worldofjoni, line 267: add `.trim()`

backend/src/startup/config.rs
```diff
@@ -232,25 +232,50 @@ impl ConfigReader {
...
+        "true" => Ok(true),
```
worldofjoni, line 268: `.parse()` does this

backend/src/startup/config.rs
```diff
@@ -232,25 +232,50 @@ impl ConfigReader {
...
+        } else {
+            info!("Google safe search feature is disabled");
+        }
+        if read_var_to_bool("USE_GEMINI_API")? {
+            info.use_gemini_api = true;
+            info.gemini_api_key = Some(read_var("GEMINI_API_KEY")?);
+            info.gemini_text_request = Some(read_var("GEMINI_TEXT_REQUEST")?);
+            info!("Using google gemini ai for image verification");
+        } else {
+            info!("Google gemini api is disabled");
```
worldofjoni, line 256: useful messages, but maybe make the wording a bit more consistent between the both.

worldofjoni (commented):
Some early feedback. I could not resist and looked at all of it 🙃.

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
```diff
@@ -0,0 +1,54 @@
...
+                Some(part) => Ok(part.text),
+            },
+        }
+    }
+
+    async fn request_api(&self, b64_image: String) -> Result<GeminiResponseJson> {
+        let resp = reqwest::Client::new()
+            .post(format!("{API_REST_URL}?key={key}", key = &self.api_key))
+            .header(reqwest::header::CONTENT_TYPE, CONTENT_TYPE)
+            .body(build_request_body(&self.text_request, &b64_image))
```
worldofjoni, line 40: `.json()` uses a json serializable argumend directly and sets the content type automatically.

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
```diff
@@ -0,0 +1,54 @@
...
+        // TODO retry with error json if response could not be decoded.
+        // TODO For now, this decode error (containing the response error json)..
+        // TODO ..will be displayed as decode error and not as api error.
+        Ok(resp.json::<GeminiResponseJson>().await?)
+    }
+}
+
+fn build_request_body(text_request: &str, b64_image: &str) -> String {
+    format!(
+        r#"{{"contents":[{{"parts":[{{"text":"{text_request}"}},{{"inline_data": {{"mime_type":"{REQUEST_TYPE}","data":"{b64_image}"}}}}]}}]}}"#
```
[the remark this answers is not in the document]
Whatsuup: That seems to be a proper solution if we consider the look, but I don't need to create a json with a new library to parse it to string an instruction later.

backend/src/layer/data/image_validation/google_api_handler.rs
Whatsuup: Even if we do that, the image evaluation is no longer at this layer and I think it should be.

backend/src/layer/data/image_validation/google_api_handler.rs
Whatsuup: I don't know a proper solution rn

worldofjoni: What do we want to do here? Right now we query both APIs even if the first fails. I think it is ok to do them "sequencially", calling the second only if the first succeeds: Safe search first, gemini second. Then you could just use `?` on the options, right?

If you really want to do them both, you still only return one error in this version, even if both fail.

Whatsuup:
This pr should be complete besides the open comments.
We should consider logging on some places.

.github/workflows/rust.yml
```diff
@@ -18,9 +18,13 @@ env:
   SMTP_USERNAME: ${{ vars.SMTP_USERNAME }}
   SMTP_PASSWORD: ${{ secrets.SMTP_PASSWORD }}
   ADMIN_EMAIL: ${{ vars.ADMIN_EMAIL }}
+  USE_SAFE_SEARCH: "true"
   GOOGLE_CLIENT_JSON_BASE64: ${{ secrets.GOOGLE_CLIENT_JSON_BASE64 }}
   GOOGLE_PROJECT_ID: ${{ vars.GOOGLE_PROJECT_ID }}
-
+  USE_GEMINI_API: "true"
+  GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
+  GEMINI_TEXT_REQUEST: "Is_a_meal_from_a_mensaria_visible_in_the_picture?"
```
Whatsuup, line 26: TODO check .env library or consider " " even though I don't think it worked

backend/.env.example
```diff
@@ -15,13 +15,22 @@ ADMIN_EMAIL=
...
 GOOGLE_PROJECT_ID=PROJECT_ID
 # Order:    adult,spoof,medical,violence,racy
 # Mapping:  0=Unknown,1=VeryUnlikely,2=Unlikely,3=Possible,4=Likely,5=VeryLikely
 IMAGE_ACCEPTANCE_VALUES=0,0,0,0,0
 
+# If you want to use the Gemini-AI for image validation, set this to true. We suggest to enable at least one of the both features.
+# GEMINI_API_KEY and GEMINI_TEXT_REQUEST needs to be set if gemini is enabled.
+USE_GEMINI_API=true
+GEMINI_API_KEY=key
+GEMINI_TEXT_REQUEST=Is_a_meal_from_a_mensaria_visible_in_the_picture?
```
Whatsuup, line 32: TODO check .env library or consider " " even though I don't think it worked

backend/README.md
```diff
@@ -62,9 +62,13 @@ The following options are available:
...
+| `GEMINI_API_KEY`                         | The Gemini API key to authenticate each api request.                                                                                                                                                                                                                                          | required, if gemini is enabled                                                                                               |
+| `GEMINI_TEXT_REQUEST`                    | The question Gemini will answer for each image. For example: Can you see food on this image? This question controls which images will be rejected or accepted.                                                                                                                                | required, if gemini is enabled                                                                                               |
```
Whatsuup, line 71: Default Question?

backend/src/layer/data/image_validation/safe_search_validation/safe_search_request.rs
```diff
@@ -72,42 +75,74 @@ impl ApiRequest {
             .header(PROJECT_ID_HEADER, &self.google_project_id)
             .header(reqwest::header::CONTENT_TYPE, CONTENT_TYPE)
             .header(reqwest::header::ACCEPT_CHARSET, CHARSET)
-            .body(build_request_body(&b64_image))
+            .body(build_request_body(b64_image).to_string())
+            // JsonValue cannot be serialised by serde as it does not implement serialise...
```
Whatsuup, line 79: see build_request_body: using the "New" json lib doesnt help that much, as serde cannot deserialize these structs (JsonValue)

backend/src/startup/config.rs
```diff
@@ -232,25 +234,54 @@ impl ConfigReader {
...
+                project_id: project_id.clone(),
+            });
+            info!("Using google safe search for image verification with cloud project '{project_id}' and category levels '{acceptance:?}'");
+        } else {
+            info!("Google safe search api is disabled");
+        }
+        if read_var_to_bool("USE_GEMINI_API").unwrap_or(DEFAULT_USE_GEMINI) {
+            info.gemini_info = Some(GeminiInfo {
+                gemini_api_key: read_var("GEMINI_API_KEY")?,
+                gemini_text_request: read_var("GEMINI_TEXT_REQUEST")?.replace('_', " "),
```
Whatsuup, line 256: TODO check .env library or consider " " even though I don't think it worked

backend/.env.example
worldofjoni: my quick tests resluted in quotes working...

backend/Cargo.toml
```diff
@@ -70,13 +70,14 @@ multer = { version = "3.1", features = ["tokio-io"] }
 minijinja = "2.0.2"
 axum-extra = { version = "0.10", features = ["typed-header"] }
 async-once-cell = "0.5.4"
+json = "0.12.4"
```
worldofjoni, line 73: not required, serde_json has its own makro

backend/README.md
worldofjoni: I would add what we end up using (initially). (not sure wheter my mail worked)

backend/src/interface/image_validation.rs
```diff
@@ -22,22 +26,25 @@ pub trait ImageValidation: Send + Sync {
...
     /// Image could not be decoded
     #[error("The provided image could not be decoded to base64: {0}")]
     ImageEncodeFailed(String),
+    /// The Image contains invalid content
+    #[error("The provided image does contain invalid content: {0}")]
+    GeminiRejectionError(String),
     /// An api related error. Returns the error provided by the api.
-    #[error("The api responded with error '{0}'.")]
-    ApiResponseError(String),
+    #[error("the api responded with an error: {1} ({0}), {2}")]
```
worldofjoni, line 43: "**T**he", maybe specify which api as there are multiple now?

backend/src/interface/image_validation.rs
```diff
@@ -22,22 +26,25 @@ pub trait ImageValidation: Send + Sync {
...
+    ApiResponseError(u32, String, String),
+    /// The response could not be determined or decoded.
+    #[error("The api responded with errors.")]
+    InvalidApiResponse,
```
worldofjoni, line 47: Differentiate better to error above?

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
```diff
@@ -0,0 +1,166 @@
+use crate::interface::image_validation::ImageValidationError::InvalidApiResponse;
+use crate::interface::image_validation::{parse_request, Result};
+use crate::layer::data::image_validation::gemini_validation::json_request::GeminiResponseJson;
+use json::JsonValue;
+
+const API_REST_URL: &str =
+    "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash:generateContent";
```
worldofjoni, line 7: We might need to keep an eye on when this api changes/depricates as it is `v1beta`.

backend/src/layer/data/image_validation/gemini_validation/gemini_evaluation.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// `text_response`<br>
+    /// This param contains all the mentioned response.
+    /// # Errors
+    /// This method returns an error if the evaluation decides to deny the image or the response could not be determined.
+    /// # Return
+    /// Nothing, what means the evaluation decided to accept the image.
+    pub fn evaluate(&self, text_response: &str) -> Result<()> {
+        if text_response.contains(ACCEPT_KEYWORD) {
+            Ok(())
+        } else if text_response.contains(REJECT_KEYWORD) {
```
worldofjoni, line 25: Would `.starts_with` be safer (as maybe the answer could contain Yes/No also later on?

backend/src/layer/data/image_validation/gemini_validation/gemini_evaluation.rs
```diff
@@ -0,0 +1,78 @@
...
+impl GeminiEvaluation {
+    /// This method evaluates the provided text provided by the gemini api.
+    /// # Params
+    /// `text_response`<br>
+    /// This param contains all the mentioned response.
+    /// # Errors
+    /// This method returns an error if the evaluation decides to deny the image or the response could not be determined.
+    /// # Return
+    /// Nothing, what means the evaluation decided to accept the image.
+    pub fn evaluate(&self, text_response: &str) -> Result<()> {
```
worldofjoni, line 22: Do we need to ignore case (upper/lower), or is this reliable enough?

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
```diff
@@ -0,0 +1,166 @@
...
+            },
+        }
+    }
+
+    async fn request_api(&self, b64_image: &str) -> Result<GeminiResponseJson> {
+        let resp = reqwest::Client::new()
+            .post(API_REST_URL)
+            .query(&[("key", &self.api_key)])
+            .body(build_request_body(&self.text_request, b64_image).to_string())
+            // JsonValue cannot be serialised by serde as it does not implement serialise...
```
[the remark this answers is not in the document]
worldofjoni: Use json macro from `serde_json` instead.

backend/src/layer/data/image_validation/google_api_handler.rs
```diff
@@ -18,37 +20,74 @@ use std::io::Cursor;
...
+                request: SafeSearchRequest::new(&info.service_account_info, info.project_id)?,
+            });
+        }
+        if let Some(info) = info.gemini_info {
+            handler.gemini_handler = Some(GeminiHandler {
+                evaluation: GeminiEvaluation::default(),
+                request: GeminiRequest::new(info.gemini_api_key, &info.gemini_text_request),
+            });
+        }
+        Ok(handler)
```
worldofjoni, line 65: That, again, I find rather unreadable (I don't like mut variables when not necessary😏).
```suggestion
        Ok(Self {
            safe_search_info = if let Some(info) = info.safe_search_info {Some(SafeSearchHandler {
                evaluation: SafeSearchEvaluation::new(info.acceptance),
                request: SafeSearchRequest::new(&info.service_account_info, info.project_id)?,
            } else {None},
            gemini_handler = info.gemini_handler.map(|info| GeminiHandler {
                evaluation: GeminiEvaluation::default(),
                request: GeminiRequest::new(info.gemini_api_key, &info.gemini_text_request),
            })
        })
```
Maybe the first part is possible even cleaner although of the `?`.
This makes it more explicit which variables Self contains. If we add new ones, they cannot simpy be forgotten to be initialized.

backend/src/layer/data/image_validation/mod.rs
```diff
@@ -1,10 +1,8 @@
-//! This component is responsible for image related api operations
+//! See [`google_api_handler`]
```
worldofjoni, line 1: Isn't the api_handler just one part of three in this module? Explicit brief description?

backend/src/layer/data/image_validation/safe_search_validation/safe_search_request.rs
worldofjoni: serde_json, see above

backend/src/startup/config.rs
```diff
@@ -232,25 +234,54 @@ impl ConfigReader {
...
-        let acceptance = read_acceptence_var("IMAGE_ACCEPTANCE_VALUES")?;
-        let service_account_info =
-            tokio::fs::read_to_string(read_var("SERVICE_ACCOUNT_JSON")?).await?;
-        info!("Using google cloud project '{project_id}' for image verification with the category levels '{acceptance:?}'");
-        Ok(ImageValidationInfo {
-            acceptance,
-            service_account_info,
-            project_id,
-        })
+        let mut info: ImageValidationInfo = ImageValidationInfo::default();
```
worldofjoni, line 239: 😏 see above. Remember that `if` are expressions, they can return a value. You can store each optional config in a variable and then cerate the `ImageValidationInfo` in the end.

backend/src/startup/config.rs
```diff
@@ -232,25 +234,54 @@ impl ConfigReader {
...
+            info!("Using google safe search for image verification with cloud project '{project_id}' and category levels '{acceptance:?}'");
+        } else {
+            info!("Google safe search api is disabled");
+        }
+        if read_var_to_bool("USE_GEMINI_API").unwrap_or(DEFAULT_USE_GEMINI) {
+            info.gemini_info = Some(GeminiInfo {
+                gemini_api_key: read_var("GEMINI_API_KEY")?,
+                gemini_text_request: read_var("GEMINI_TEXT_REQUEST")?.replace('_', " "),
+            });
+            info!("Using google gemini api for image verification");
```
worldofjoni, line 258: maybe print used query sting?

worldofjoni (approved):
Thanks for getting this done so quickly. As usual, some minor nit picks 😉 .

worldofjoni:
As for logging, I thing we don't need to add any aditional. Errors are passed up. We don't do any "independent" actions.

worldofjoni:
One thing that is still missing AFAIK:
- [ ] finding / evaluating the propper query question and ensuring it satisfies our needs (low false positives). Maybe we could automate that by creating a test that queries a folder of images, compares them against the required answer and creates a short report. (that test can be `#[ignore]`ed for automatic testing). This way, we can easily compare different query questions with the press of a button.

worldofjoni (approved):

worldofjoni:
> Correct images rejected: 2/19
> Incorrect images accepted: 17/19

More interesting would be the actual false positive/negative rates.
https://en.wikipedia.org/wiki/False_positive_rate

Whatsuup:
New output prompt contains FPR as well as FNR. (FPR images can be manually deleted via review, FNR images are lost in the process.)

```
--- Results: ---
Correct decisions: 110/129 (85.27132%).
Wrong decisions: 19/129 (14.728682%).

Images that got accepted but should not (FP): 18/129.
False positive rate (FPR): 47.368423%.

Images that got not accepted but should be (FN): 1/129.
False negative rate (FNR): 1.0989012%.
----------------
```

PR #206, opened 2026-04-23
Bump dependencies

worldofjoni:


worldofjoni:
@Whatsuup do you have an idea why these geminy API requests fail? Does that have something to do with the deprication of the specific model we use?

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
```diff
@@ -4,9 +4,10 @@ use crate::layer::data::image_validation::gemini_validation::json_request::Gemin
 
 // Consider: This is the beta version. Change if depreciated!
 const API_REST_URL: &str =
-    "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash:generateContent";
+    "https://generativelanguage.googleapis.com/v1beta/models/gemini-flash-latest:generateContent";
 const REQUEST_TYPE: &str = "image/jpeg";
-const REQUEST_SPECIFICATION: &str = "Answer yes or no and give a short explanation in English.";
+const REQUEST_SPECIFICATION: &str =
+    "Answer yes or no and give a short explanation in English. In case of doubt, answer with yes.";
```
Whatsuup: Why?

Whatsuup (approved):

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
worldofjoni: I had that in my history. I think we once thought about that to promote not rejecting images. We can revert this (to also just see the changes of the new model). What do you think?

worldofjoni:
@Whatsuup did you check wether there are changes in API billing?

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
Whatsuup: We can keep it like that, but I guess a stricter prompt would be more suitable in this case. Instead of "answer with yes" -> "answer with no". But I dont know how this influences the decision at all. 

We can leave it like this and look at the reports more often.

Whatsuup:
> @Whatsuup did you check wether there are changes in API billing?

As this change only affects the model version, no extra costs are expected.

We can consider to use a lite version, which does not support complex reasoning but is much faster and "cheaper". 
If then a decision is not decidable we can build a two stage check, first the lite model than the smarter model. Maybe in the future, because the costs are not that high...

backend/src/layer/data/image_validation/gemini_validation/gemini_request.rs
worldofjoni: we back then thought about accepting unclear images, so we can review them. I will revert this for now.

